PR #4, opened 2021-05-26
[FIX] theme_avantgarde: fix duplicated theme_avantgarde.s_cover view

smottet:


theme_avantgarde/views/customizations.xml
```diff
@@ -28,7 +28,7 @@
 </template>
 
 <!-- ======== BANNER ======== -->
- <template id="s_cover" inherit_id="website.s_cover" name="Avantgarde s_cover">
```
qsm-odoo, line 31: Meh... that fail :|

Did you speak with the designers to know it was supposed to be banner here? What actually happened with two s_cover defined? The first one was simply ignored? If yes, should not it be simply removed since nobody actually saw there was a mistake? :grimacing:

qsm-odoo (changes requested):
Commit tile should start with a verb + missing task reference

theme_avantgarde/views/customizations.xml
stefanorigano: Hello @smottet , it was supposed to be `s_banner`, thanks.
Sorry for the headache.

qsm-odoo (changes requested):
Is the fix required in 14.0 for anything? Otherwise retarget for master (or 14.3 if you want) to avoid breaking potential custo.

+ task-XXX

smottet:
> Is the fix required in 14.0 for anything? Otherwise retarget for master (or 14.3 if you want) to avoid breaking potential custo.
> 
> * task-XXX

No particular reason. I will re-target master then.

JKE-be:
@robodoo r+

PR #5, opened 2021-05-27
[IMP] website: change CTA for schedule appointment objective

smottet:
If user selects 'schedule appointments' as main objective in the configurator
and if the website_calendar module is installed then the Call To Action of
snippets s_banner, s_cover and s_call_to_action is changed to 'Schedule an
appointment' and redirect user to '/calendar' on click.


JKE-be:
@robodoo r+

PR #6, opened 2021-05-28
[FIX] website_animate: fix animations rotate in down

bvr-odoo:
Rotate-Down Animations did not launch correctly because we were trying
to get the position of the element (by not taking CSS transforms into
calculations) via the matrix of the css transform and it doesnt work
if there is a transform-origin on the element.

Now we find the top position by checking that of its parent.

task-2215118

### DIFF
--- a/website_animate/static/src/js/o_animate.frontend.js
+++ b/website_animate/static/src/js/o_animate.frontend.js
@@ -53,10 +53,8 @@ var WebsiteAnimate = {
                 var state     = $el.css("animation-play-state");
 
                 // We need to offset for the change in position from some animation
-                // So we get the top value of the transform matrix
-                var transformMatrix = $el.css('transform').replace(/[^0-9\-.,]/g, '').split(',');
-                var transformOffset = transformMatrix[13] || transformMatrix[5];
-                var elTop = $el.offset().top - transformOffset;
+                // So we get the top value by not taking CSS transforms into calculations
+                var elTop = self.getElementOffsetTop($el[0]);
 
                 var visible = windowBottom > (elTop + elOffset) && windowTop < (elTop + elHeight - elOffset);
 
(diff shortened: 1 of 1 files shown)


website_animate/static/src/js/o_animate.frontend.js
```diff
@@ -105,6 +103,21 @@ var WebsiteAnimate = {
...
         });
     },
+
+    // Get element top offset by not taking CSS transforms into calculations
+    getElementOffsetTop: function (el) {
+        let top = 0;
+        let element = el;
+      
+        // Loop through the DOM tree
+        // and add it's parent's offset to get page offset
```
qsm-odoo, line 113: - "its"
- Break line at 80 characters (not 30)

website_animate/static/src/js/o_animate.frontend.js
```diff
@@ -105,6 +103,21 @@ var WebsiteAnimate = {
             });
         });
     },
+
+    // Get element top offset by not taking CSS transforms into calculations
+    getElementOffsetTop: function (el) {
+        let top = 0;
+        let element = el;
```
qsm-odoo, line 110: Don't use `let` in 12.0

qsm-odoo (changes requested):
How can I reproduce the bug you are fixing here?

bvr-odoo:
> How can I reproduce the bug you are fixing here?

The animations "Rotate In-Down-Left" and "Rotate In-Down-Right" doesnt work without this fix.

qsm-odoo:
> > How can I reproduce the bug you are fixing here?
> 
> The animations "Rotate In-Down-Left" and "Rotate In-Down-Right" doesnt work without this fix.

I would not ask that question if that was that simple to reproduce... :/ I just retested in 12.0 on both Chrome and Firefox and there is no problem on my side :grimacing: I can't see the difference with or without your fix, so please give me more info :wink:

bvr-odoo:
> > > How can I reproduce the bug you are fixing here?
> > 
> > 
> > The animations "Rotate In-Down-Left" and "Rotate In-Down-Right" doesnt work without this fix.
> 
> I would not ask that question if that was that simple to reproduce... :/ I just retested in 12.0 on both Chrome and Firefox and there is no problem on my side  I can't see the difference with or without your fix, so please give me more info

I'm sorry but I don't really have more info to give you :)
I have the issue in V12 in Chrome and in Firefox

1 - drop a snippet text-image
2 - add Rotate In-Down-Left animation on this snippet
3 - save the page
4 - the animation is not started

qsm-odoo (approved):
@robodoo r+

PR #7, opened 2021-05-28
[FIX] website_animate: fix drag n drop an animated snippet

bvr-odoo:
Before this commit, when drag n drop a snippet to move it in a website
page. The snippet was invisible after the drop.

task-2215118

### DIFF
--- a/website_animate/static/src/js/o_animate.editor.js
+++ b/website_animate/static/src/js/o_animate.editor.js
@@ -21,25 +21,6 @@ sOptions.registry.o_animate = sOptions.Class.extend({
     onFocus: function () {
         this._setActive(); // Needed as in charge of hiding duration/delay/...
     },
-    /**
-     * @override
-     */
-    cleanForSave: function () {
-        // Clean elements
-        this.$target.removeClass('o_animating o_animated o_animate_preview')
-                    .css({
-                        'animation': '',
-                        'animation-name': '',
-                        'animation-play-state': '',
-                        'visibility': '',
-                    });
-        if (this.$target.hasClass('o_animate')) {
-            this.$target.css('animation-play-state', 'paused');
-        }
-
-        // Clean all inView elements
-        $('#wrapwrap').find('.o_animate').removeClass('o_visible');
-    },
 
     //--------------------------------------------------------------------------
     // Options
--- a/website_animate/static/src/js/o_animate.frontend.js
+++ b/website_animate/static/src/js/o_animate.frontend.js
@@ -126,6 +126,19 @@ publicWidget.registry.WebsiteAnimate = publicWidget.Widget.extend({
 
         return this._super.apply(this, arguments);
     },
+    /**
+     * @override
+     */
+    destroy: function () {
+        this._super.apply(this, arguments);
+        this.$target.find('.o_animate')
+            .removeClass('o_animating o_animated o_animate_preview')
+            .css({
+                'animation-name': '',
+                'animation-play-state': '',
+                'visibility': '',
+            });
+    },
 });
 
 // Backward compatibility for enark animation system


website_animate/static/src/js/o_animate.editor.js
```diff
@@ -34,7 +34,7 @@ sOptions.registry.o_animate = sOptions.Class.extend({
                         'visibility': '',
                     });
         if (this.$target.hasClass('o_animate')) {
-            this.$target.css('animation-play-state', 'paused');
```
bvr-odoo, line 37: I dont know why this line was usefull.. this property is anyway settled with 'paused' in the start method of o_animate.frontend.js

website_animate/static/src/js/o_animate.editor.js
```diff
@@ -34,7 +34,7 @@ sOptions.registry.o_animate = sOptions.Class.extend({
...
+            this.$target.css('visibility', 'visible');
```
qsm-odoo, line 37: I doubt this is the right fix. The o_animate class sets `visibility: hidden` but once the snippet is saved we cancel that via inline style? :thinking: :thinking:

website_animate/static/src/js/o_animate.editor.js
qsm-odoo: There may be no reason at all, that's why we need to review the entire app in master :wink:

qsm-odoo (changes requested):
Commit message:

- "Before this commit, when drag n drop a snippet to move it in a website
page. The snippet was invisible after the drop." -> probably a comma instead of a point :wink: 

- "We have also removed the line setting the animation-play-state with
'paused' as this also causes an issue with the drag and drop. And this
line does not seem usefull". -> don't explain the code in commit message, explain what it fixes

website_animate/static/src/js/o_animate.editor.js
```diff
@@ -51,7 +32,15 @@ sOptions.registry.o_animate = sOptions.Class.extend({
     selectClass: function (previewMode, value, $opt) {
         this._super.apply(this, arguments);
         forceAnimation(this.$target);
-        this.$target.toggleClass('o_animate_preview o_animate', !!value);
+        this.$target.toggleClass('o_animate o_animate_preview', !!value);
```
qsm-odoo, line 35: Why that change?

website_animate/static/src/js/o_animate.editor.js
```diff
@@ -51,7 +32,15 @@ sOptions.registry.o_animate = sOptions.Class.extend({
...
+    },
+    /**
+     * @override
+     */
+     destroy: function () {
+        this._super.apply(this, arguments);
+        // Clean elements
+        this.$target.removeClass('o_animate_preview');
```
qsm-odoo, line 43: You cannot modify the target in the `destroy` of an option, why not in `cleanForSave` (which is normally called around the same time as `destroy` anyway)?

website_animate/static/src/js/o_animate.frontend.js
```diff
@@ -117,6 +117,19 @@ publicWidget.registry.WebsiteAnimate = publicWidget.Widget.extend({
...
+     */
+    destroy: function () {
+        this._super.apply(this, arguments);
+        // Clean elements
+        this.$target.find('.o_animate').removeClass('o_animating o_animated')
+                    .css({
+                        'animation-name': '',
+                        'animation-play-state': '',
+                        'visibility': '',
+                    });
```
qsm-odoo, line 131: Use a 4-space indentation instead of aligning with `find`

website_animate/static/src/js/o_animate.frontend.js
```diff
@@ -117,6 +117,19 @@ publicWidget.registry.WebsiteAnimate = publicWidget.Widget.extend({
 
         return this._super.apply(this, arguments);
     },
+    /**
+     * @override
+     */
+    destroy: function () {
+        this._super.apply(this, arguments);
+        // Clean elements
```
qsm-odoo, line 125: Remove this useless comment

qsm-odoo (changes requested):

website_animate/static/src/js/o_animate.editor.js
bvr-odoo: for nothing, (mishandling)

website_animate/static/src/js/o_animate.editor.js
bvr-odoo: Indeed, Actually I have to remove `this._super.apply(this, arguments);`  from the cleanForSave to fix the bug

website_animate/static/src/js/o_animate.frontend.js
```diff
@@ -126,6 +126,18 @@ publicWidget.registry.WebsiteAnimate = publicWidget.Widget.extend({
...
+     * @override
+     */
+    destroy: function () {
+        this._super.apply(this, arguments);
+        this.$target.find('.o_animate').removeClass('o_animating o_animated')
+                        .css({
+                            'animation-name': '',
+                            'animation-play-state': '',
+                            'visibility': '',
+                        });
```
qsm-odoo, line 139: @bvr-odoo I did not mean 4 space indentation after `find` but after `this`

qsm-odoo (changes requested):
I tested the branch, it does seem to fix the bug on my side, can you retest?


https://user-images.githubusercontent.com/10338094/127483315-e8b1465c-0364-42fb-9dd6-0f1fe8c5061c.mp4

bvr-odoo:
> I tested the branch, it does seem to fix the bug on my side, can you retest?
> 
>  bug_dd_animate.mp4

Indeed,  I don't understand what i missed...
Actually the clean for save method is called on snippet drop so "o_animate_preview" is removed, that's why the snippet is invisible.

website_animate/static/src/js/o_animate.frontend.js
```diff
@@ -126,6 +126,19 @@ publicWidget.registry.WebsiteAnimate = publicWidget.Widget.extend({
...
+    destroy: function () {
+        this._super.apply(this, arguments);
+        this.$target.find('.o_animate')
+            .removeClass('o_animating o_animated o_animate_preview')
+            .css({
+                'animation-name': '',
+                'animation-play-state': '',
+                'visibility': '',
+            });
+    },
```
bvr-odoo, line 141: because the clean for save methos is called on snippet drop (in the page), I had to move the remove of the "o_animate_preview" class here. This class is specific to the edit mode but I didnt find an other solution

qsm-odoo: It's fine removing it for now, it's a particular case. But could you not *add* it in the **widget** start too? instead of via the options? That way it would be consistent at least: in the widget start, you could try `if (editableMode) { ... add previewClass ... }` and then it's normal that it is removed in its `destroy`. And not mention in the editor option at all anymore.

bvr-odoo: The widget is on the wrapwrap in the .js frontend so it won't work because the start is only called once.
We need to add the class "o_animate_preview" each time we add the class "o_animate" with the options.

qsm-odoo (approved):
@robodoo r+

PR #9, opened 2021-05-28
[ADD] theme_aviato: Add new theme Aviato

xlu-odoo:
Add a travel theme with illustrations, imagery and shapes for WB
Reviewer : @Cocographique


Cocographique:
You can remove the SVG files from the /shapes folder.

theme_aviato/views/snippets/s_picture.xml
```diff
@@ -0,0 +1,26 @@
...
+<odoo>
+    <template id="s_picture" inherit_id="website.s_picture">
+        <xpath expr="//section" position="attributes">
+            <attribute name="class" add="pt72 overflow-hidden o_cc3" remove="pt48 o_cc2" separator=" "/>
+            <attribute name="data-oe-shape-data">{"shape":"web_editor/Bold/01","flip":[]}</attribute>
+        </xpath>
+
+        <xpath expr="//section/div" position="before">
+            <div class="o_we_shape o_web_editor_Bold_01 bg-o-color-3"/>
+        </xpath>
```
qsm-odoo, line 11: This is good (adding o_cc3 as preset, forcing the bold shape to use the same bg color as that preset so that the text renders well over the shape, then add a bg color on the shape itself) :+1:

However, it may be not perfect in the new system: when dropping that snippet, the user has to understand that to control the background color of the snippet, it has to first remove the shape / reset it background transparency :grimacing: Maybe it is best to leave the shape background transparent when we can: now you can add that `bg-o-color-3` class on the snippet itself **alongside the o_cc3 class**. That will have the same visual aspect of your snippet here but the editor panel status and further edition will make more sense IMO.

theme_aviato/__manifest__.py
```diff
@@ -0,0 +1,36 @@
+{
+    'name': 'Aviato',
```
qsm-odoo, line 2: Find the intruder ;)
Screenshot from 2021-08-30 11-58-50

qsm-odoo (changes requested):
@xlu-odoo Real-quick review here, my goal was to give quick pointers on mistakes to avoid: the plan is for me to merge themes on Friday / Monday (while we continue to merge new features this week :grimacing:). Meanwhile, the themes should be made perfect (hello `master-perfect-themes-qsm` branch of last year :rofl:), so maybe share this review to your fellow designer colleagues :) cc @Cocographique 

Other things to consider, that I could not comment inline: the images are too heavy here. From what I remember, other themes have images optimized around 200kb max, most of the time even around 100kb).

PR #11, opened 2021-05-28
[FIX] theme_*: customize CTA for all navigation bar templates

oussamamessaoudi:
task-2368576

### DIFF
--- a/theme_cobalt/views/customizations.xml
+++ b/theme_cobalt/views/customizations.xml
@@ -5,14 +5,8 @@
 
 <!-- ==== Headers ===== -->
 <template id="header_shadow" inherit_id="website.template_header_default" name="Cobalt Header Shadow">
-    <xpath expr="//nav" position="attributes">
-        <attribute name="class" remove="shadow-sm" separator=" "/>
-    </xpath>
-</template>
-
-<template id="header_button" inherit_id="website.template_header_default_oe_structure_header_default_1" name="Cobalt Header Button">
-    <xpath expr="//a[hasclass('btn')]" position="attributes">
-        <attribute name="class" add="btn-secondary" remove="btn-primary" separator=" "/>
+    <xpath expr="//t[@t-set='_navbar_classes']" position="attributes">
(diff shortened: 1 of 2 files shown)


website_animate/views/options.xml
```diff
@@ -16,50 +16,59 @@
...
-                    <we-divider/>
-                    <we-button data-select-class="o_anim_flip_in_x">Flip-In-X</we-button>
-                    <we-button data-select-class="o_anim_flip_in_y">Flip-In-Y</we-button>
-                </we-select>
-                <we-input string="Animation Duration" data-dependencies="!no_animation_opt"
-                          data-select-style="0.4s" data-css-property="animation-duration" data-unit="s"/>
-                <we-input string="Animation Delay" data-dependencies="!no_animation_opt"
-                          data-select-style="0s" data-css-property="animation-delay" data-unit="s"/>
-                <we-checkbox string="Each time it becomes visible" data-dependencies="!no_animation_opt"
-                             data-select-class="o_animate_both_scroll"/>
```
qsm-odoo, line 61: What are those changes?

qsm-odoo (changes requested):

website_animate/views/options.xml
oussamamessaoudi: Adding  `t-att-data-display-handles="handlesDisplay"` to avoid displaying 2 overlays
6819052-9fbf32097aec9f3efecb0476cb0d8722

qsm-odoo: I canceled your PR about "data-display-handles", that should not be necessary and I don't see why there would 2 trashes icons on a rebased version of master:
Screenshot from 2021-06-18 14-22-36

Please re-investigate why there is this bug *in your branch*

oussamamessaoudi: it is not a bug in my branch, it was a bug in master and i think it is fixed.

qsm-odoo (approved):
@robodoo r+

qsm-odoo:
@robodoo r+

PR #17, opened 2021-06-10
[IMP] web_editor: add animated shapes

Brieuc-brd:
The purpose is to update the shapes of themes depending on the PR [#70843.](https://github.com/odoo/odoo/pull/70843)
Also, it set correctly the colors of the shapes that use the colors of the header or the footer.

Task-2319488

### DIFF
--- a/theme_artists/views/snippets/s_call_to_action.xml
+++ b/theme_artists/views/snippets/s_call_to_action.xml
@@ -4,11 +4,11 @@
         <!-- Section -->
         <xpath expr="//section" position="attributes">
             <attribute name="class" add="pb152 pt64" remove="pb24 pt48" separator=" "/>
-            <attribute name="data-oe-shape-data">{"shape":"web_editor/Airy/12","flip":["y"]}</attribute>
+            <attribute name="data-oe-shape-data">{"shape":"web_editor/Airy/12_001","flip":["y"]}</attribute>
         </xpath>
         <!-- Shape -->
         <xpath expr="//div[hasclass('container')]" position="before">
-            <div class="o_we_shape o_web_editor_Airy_12 o_we_flip_y"/>
+            <div class="o_we_shape o_web_editor_Airy_12_001 o_we_flip_y"/>
         </xpath>
         <!-- Paragraph -->
         <xpath expr="//h3" position="replace">
--- a/theme_paptic/static/src/scss/primary_variables.scss
+++ b/theme_paptic/static/src/scss/primary_variables.scss
@@ -87,3 +87,4 @@ $o-color-palettes-compatibility-indexes: (
 $o-bg-shapes: change-shape-colors-mapping('web_editor', 'Blocks/04', (3: 1));
 $o-bg-shapes: change-shape-colors-mapping('web_editor', 'Origins/03', (3: 1));
 $o-bg-shapes: change-shape-colors-mapping('web_editor', 'Origins/01', (2: 3, 5: 3));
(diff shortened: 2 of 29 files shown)


theme_artists/views/snippets/s_call_to_action.xml
```diff
@@ -8,7 +8,7 @@
         </xpath>
         <!-- Shape -->
         <xpath expr="//div[hasclass('container')]" position="before">
-            <div class="o_we_shape o_web_editor_Airy_12 o_we_flip_y"/>
+            <div class="o_we_shape o_web_editor_Airy_12 o_we_flip_y o_footer_extra_shape_mapping"/>
```
qsm-odoo: This should not be done, the point is that this class is automatically added when building pages **through the new website configurator**. When using the editor, you don't want that class when dropping a snippet into the page, otherwise the shape would be linked to the footer color wherever it is dropped.

theme_paptic/static/src/scss/primary_variables.scss
```diff
@@ -84,6 +84,6 @@ $o-color-palettes-compatibility-indexes: (
 // Shapes
 //------------------------------------------------------------------------------
 
-$o-bg-shapes: change-shape-colors-mapping('web_editor', 'Blocks/04', (3: 1));
+$o-bg-shapes: change-shape-colors-mapping('web_editor', 'Blocks/04', (3: 1, 1: 2));
```
qsm-odoo, line 87: Should this be done in this commit? I don't see Blocks/04 changed in the community PR, is this a fix? If yes, it should be done in another PR.

theme_paptic/views/customizations.xml
```diff
@@ -206,11 +206,11 @@
 <!-- ==== Three Columns ===== -->
 <template id="s_three_columns" inherit_id="website.s_three_columns" name="Paptic s_three_columns">
     <xpath expr="//section" position="attributes">
-        <attribute name="class" remove="bg-200 pt32 pb32 o_cc o_cc3" add="pt120 pb120" separator=" "/>
-        <attribute name="data-oe-shape-data">{"shape":"web_editor/Origins/01","flip":[]}</attribute>
+        <attribute name="class" remove="bg-200 pt32 pb32" add="pt120 pb120 o_cc1 bg-o-color-2" separator=" "/>
```
qsm-odoo, line 209: This will be in conflict with another task we have already fixing/chaning that. But for the record, what you do here is not ok: you cannot have a o_ccX class and a bg-o-color-X class at the same time, so remove this change :wink:

qsm-odoo (changes requested):

theme_artists/views/snippets/s_call_to_action.xml
Brieuc-brd: Ok I thought it was an issue, but yeah it's logical in fact :). I removed the [FIX] commit.

theme_paptic/static/src/scss/primary_variables.scss
Brieuc-brd: The render of this snippet is not the same than the screenshot (paptic_screenshot.jpg).
We'll probably update that in another PR with some other updates on this theme :). I removed the [FIX] commit.

theme_paptic/views/customizations.xml
Brieuc-brd: @qsm-odoo 
Ok I removed the change. But without that, the shape is hidden (it's white on a white background).
Should I let like that ?

qsm-odoo (approved):
@robodoo r+

PR #19, opened 2021-06-16
[IMP] theme_kiddo: theme design improvement

karlosvalverde:
temp

--

Task-2573229


theme_kiddo/models/theme_kiddo.py
```diff
@@ -6,10 +6,10 @@ class ThemeKiddo(models.AbstractModel):
 
     def _theme_kiddo_post_copy(self, mod):
         self.disable_view('website.template_header_default')
-        self.enable_view('website.template_header_centered_logo')
+        self.enable_view('website.template_header_default_align_right')
```
Cocographique, line 9: There is something wrong here.
`template_header_default_align_right` is an option of the default header template.
But as `template_header_default` is disabled the line before, the menu from the portal module is shown.

Screenshot 2021-09-23 at 16 44 10

theme_kiddo/views/snippets/s_banner.xml
```diff
@@ -1,23 +1,37 @@
...
+    </xpath>
+
+    <!-- Paragraph -->
+    <xpath expr="//div[hasclass('col-lg-6')]" position="replace">
+        <div class="col-lg-6 pt48 pb48 m-3 m-md-0 p-4 p-md-5 o_colored_level o_cc o_cc3 rounded shadow border" style="border-radius: 35px !important; box-shadow: rgb(104, 70, 114) 0px 0px 35px 0px !important; border-color: rgb(104, 70, 114) !important; border-width: 4px !important; border-style: dotted !important;">
+            <h1><font style="font-size: 62px; line-height: 48px;">A little place of paradise.</font></h1>
+            <p><br/></p>
+            <p class="lead">Because your children deserve the best, we welcome children from 0 to 3 years in a warm and specific environment for the needs of small children</p>
+            <p><br/></p>
+            <a t-att-href="cta_btn_href" class="btn btn-primary mb-2"><t t-esc="cta_btn_text">About us</t></a>
```
Cocographique, line 32: If you just change the text, you can use the `mode="inner"` here.

theme_kiddo/models/theme_kiddo.py
qsm-odoo: Yeah... the align_right stuff is not important but if you disable all header template... well there is no menu :grimacing:

PR #22, opened 2021-06-25
[IMP] theme_loftspace: update the theme

Brieuc-brd:
This commit update the theme to improve the design with shapes, new colors, new pictures and new custom snippets.

**Improvements :**
- Change the header
- Change the tour
- Change the color palette

**Snippets improved :**
- Call_to_action
- Cover
- Text_image
- Three_columns
- Title

**New customisations :**
- Image_gallery
- Banner
- Image_text
- Numbers
- Color_blocks_2
- Features
- Media_list
- Features_grid
- Comparisons
- Product_catalog

Remove outdated theme description.

Task-2557927


theme_loftspace/views/snippets/s_call_to_action.xml
```diff
@@ -12,14 +17,21 @@
...
+            <h3 style="font-size: 36px;">Would you like more information <br/>about a product?</h3>
         </xpath>
         <xpath expr="//p" position="replace"/>
         <!-- Button -->
         <xpath expr="//div[hasclass('col-lg-12')]/p" position="attributes">
             <attribute name="style" add="text-align: center;" remove="text-align: right;" separator=" "/>
         </xpath>
+        <xpath expr="//a[hasclass('btn')]" position="attributes">
+           <attribute name="class" remove="mb-2" separator=" "/>
+        </xpath>
```
qsm-odoo, line 31: Normally, the mb-2 class is there for a reason and should not have an impact on your snippet? (it was added as a fix in case you duplicate the button, you want a vertical margin in mobile to have a space between your two buttons).

theme_loftspace/views/snippets/s_cover.xml
```diff
@@ -2,13 +2,13 @@
...
-        <xpath expr="section" position="attributes">
-            <attribute name="class" add="o_half_screen_height" separator=" "/>
+        <xpath expr="//section" position="attributes">
+            <attribute name="class" add="o_full_screen_height pt232" remove="pt96" separator=" "/>
         </xpath>
         <!-- Title -->
         <xpath expr="//h1" position="replace">
             <h1 style="text-align: center;">
-                <font style="font-size: 62px;"><b>Bringing life to your interior</b></font>
+                <span style="font-size: 62px;"><b>Bringing life to your interior</b></span>
```
qsm-odoo, line 11: The editor still seems to use the "font" tag when editing font sizes, so I think font should be kept?

theme_loftspace/static/src/scss/primary_variables.scss
```diff
@@ -121,18 +121,27 @@ $o-theme-font-configs: (
 // Website customizations
 //------------------------------------------------------------------------------
 
+// Change color preset for the footer
+$o-base-color-palette: map-merge($o-base-color-palette, (
+    'footer': 3,
+));
```
qsm-odoo, line 127: o-base-color-palette should normally *never* be changed by a theme. Like its name advertises, it is meant to define the default values each potential key should have, meaning that a palette not defining a value for that key would use that default value. Doing this, you basically change all footer colors of all the palettes that do not define one and that the user can choose... very random.

If your theme wants to use the preset color n°3 for its footer, it should normally use a color palette which use that for its footer... so here I see you use "graphene-2" as color palette, the normal way would be to change the graphene-2 palette for all themes unfortunately... to see if that make sense.

Brieuc-brd: After working on it, this change is not needed and the theme can work with the default color of the footer 👍 .

Brieuc-brd:
Merged here : https://github.com/odoo/design-themes/pull/500

PR #28, opened 2021-07-06
[FIX] theme_*: use correct contact header SCSS variable

qsm-odoo:
The contact header scss variable value was written with an uppercase
unlike other header scss variables. Unfortunately, fixing that typo is
not stable-friendly (as people already using the template now saved
that typo in their DB). Instead, we use that typo as default value for
the related themes.

Related to task-2368576


qsm-odoo (commented):
@robodoo r+

PR #33, opened 2021-07-07
[FIX] theme_*: check cta_btn_text is not False before applying upper

smottet:
The themes cobalt, paptic and vehicle apply upper on cta_btn_text for some
snippets. In some case cta_btn_text is a boolean (False) and not a string.
We must check cta_btn_text is not False before calling upper on it.

task-2518565


JKE-be:
@robodoo r+

b

Xavier-Do:
I think that runbot think this is linked to a 10.0 branch. Recycling pr number does not look like a good idea! :D

Xavier-Do:
Mannually fixed runbot side this should be green soon.

JKE-be:
@robodoo r+

PR #54, opened 2021-07-27
[FIX] website_animate: fix start of animations

bvr-odoo:
Before this commit, the calculation of the top position of the
animated elements did not take into account the fact that the
scrollable element is no longer the body. And therefore the animations
never started.

task-2215118


qsm-odoo (approved):
@robodoo r+

PR #55, opened 2021-07-27
[FIX] test_themes: adapt TestThemesWebsiteSelector to new registry

sdegueldre:
In odoo/odoo#72675 the registry for navbar widgets was changed to use
the new registry class introduced with the wowl webclient, this use site
was not adapted, causing a crash on startup.


qsm-odoo (approved):
@robodoo r+

PR #267, opened 2021-07-28
[IMP] theme_yes: update theme

edi-odoo:
Update the theme to improve the design with shapes, new colors, new pictures and new custom snippets.
The tour has also been adapted.

[task-2555035](https://www.odoo.com/web#id=2555035&action=333&active_id=4733&model=project.task&view_type=form&cids=1&menu_id=4720)


theme_yes/static/src/js/tour.js
```diff
@@ -8,41 +8,39 @@ var tour = require("web_tour.tour");
 
 const snippets = [
     {
-        id: 's_carousel',
-        name: 'Carousel',
+        id: 'coin',
```
AntoineVDV, line 11: 🦆

theme_yes/static/src/js/tour.js
```diff
@@ -8,41 +8,39 @@ var tour = require("web_tour.tour");
 
 const snippets = [
     {
-        id: 's_carousel',
-        name: 'Carousel',
+        id: 's_cover', // whatever the id, it works anyway
```
JKE-be, line 11: yes, it is an id :/
Purpose is to allow tour to select this Cover later and not another one if your tour insert 2 'Cover'...
it could be the index 1,2,3,4...  it will works

name is to select the snippet to drop, id to find it e.g. in selectSnippetColumn, changeIcon, changeImage, ...

So, just be sure in case you drop 2 snippets identical to use coin and bird as id ;) :bird:

PR #268, opened 2021-07-28
[FIX] theme_kiddo: fix typo in default font name

qsm-odoo:
Because of a typo, the theme used "sans-serif" as font, thus with a
random look according to OS and browser.

opw-2559735


qsm-odoo (commented):
@robodoo r+

PR #482, opened 2021-08-09
[IMP] theme_kiddo, theme_vehicule: add images_peview_theme in manifest

JKE-be:
Example of how to communicate image to be used by IAP in preview when use_theme_image is checked on iap.
//cc @Cocographique

### DIFF
--- a/theme_kiddo/__manifest__.py
+++ b/theme_kiddo/__manifest__.py
@@ -17,11 +17,24 @@
         'views/snippets/s_three_columns.xml',
         'views/snippets/s_call_to_action.xml',
 
-        ],
+    ],
     'images': [
         'static/description/Kiddo_description.png',
         'static/description/kiddo_screenshot.jpg',
     ],
+    'images_preview_theme': {
+        'website.s_banner_default_image': '/theme_kiddo/static/src/img/library/bg11.jpg',
+        'website.s_image_text_default_image': '/theme_kiddo/static/src/img/content/content_img_14.jpg',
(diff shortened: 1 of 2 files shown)


JKE-be:
@robodoo r+

//cc @Cocographique

theme_kiddo/__manifest__.py
```diff
@@ -17,11 +17,47 @@
         'views/snippets/s_three_columns.xml',
         'views/snippets/s_call_to_action.xml',
 
-        ],
+    ],
     'images': [
         'static/description/Kiddo_description.png',
         'static/description/kiddo_screenshot.jpg',
     ],
+    'images_peview_theme': {
```
qsm-odoo, line 25: "peview" ? :wink:

qsm-odoo (changes requested):

JKE-be:
@qsm-odoo thank you ;) 
don't hesitate to r- when you are sure that it is wrong ;) 
I don't bite and I'm able to see that I'm wrong when it is the case (most of the time :p) ;)

theme_vehicle/__manifest__.py
```diff
@@ -16,6 +16,34 @@
         'static/description/vehicle_description.png',
         'static/description/vehicle_screenshot.jpeg',
     ],
+    'images_peview_theme': {
```
rdeodoo, line 19: :eye:

rdeodoo: fixed this one
 I guess it can be  r+ now?

JKE-be:
@robodoo r+

PR #486, opened 2021-08-18
[FIX] theme_bistro, theme_zap: remove misleading 'hambuger' variable

qsm-odoo:
Those two themes were defining a 'hambuger-type' variable to
'off-canvas', probably a typo of 'hamburger-type'... but the themes
do not even activate the views related to 'off-canvas', so this had
simply no effect and did not create any bug.


qsm-odoo (commented):
@robodoo r+

PR #489, opened 2021-08-23
[FIX] website_animate: fix each time it becomes visible option

bvr-odoo:
Before this commit, the option "each time it becomes visible" did not
work in edit mode.

task-2630112

### DIFF
--- a/website_animate/static/src/js/o_animate.frontend.js
+++ b/website_animate/static/src/js/o_animate.frontend.js
@@ -45,7 +45,7 @@ var WebsiteAnimate = {
             var direction = (windowTop < lastScroll) ? -1 : 1;
             lastScroll = windowTop;
 
-            self.items.each(function () {
+            $(".o_animate").each(function () {
                 var $el       = $(this);
                 var elHeight  = $el.height();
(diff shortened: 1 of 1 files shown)


website_animate/static/src/js/o_animate.frontend.js
```diff
@@ -5,16 +5,14 @@ var publicWidget = require('web.public.widget');
...
     win   : {},
-    items : {},
 
     offsetRatio : 0.3,  // Dynamic offset ratio: 0.3 = (element's height/3)
     offsetMin   : 10,   // Minimum offset for small elements (in pixels)
 
     // Retrieve animable elements and attach handlers.
     start: function () {
         var self   = this;
-        self.items = $(".o_animate");
```
qsm-odoo, line 16: As a fix, keep this assignment (and its definition you removed above too) even if the variable is not used anymore. Otherwise, custo might be broken.

qsm-odoo (changes requested):

qsm-odoo (approved):
@fw-bot up to saas-14.4
@robodoo r+

PR #493, opened 2021-09-02
[FIX] theme_(vehicle|yes|zap): fix masonry xpath

smottet:
The PR odoo/odoo#67361 modifies the website.s_masonry_block template
and make it call website.s_masonry_block_default_template The xpaths
should now be done in a template inheriting from website.s_masonry_
block_default_template instead.


qsm-odoo (approved):
@robodoo r+

PR #499, opened 2021-09-23
[FIX] theme_vehicle: fix btn-lg size

xlu-odoo:
This commit removes the modified size for large (lg) buttons. They were far too big.


xlu-odoo:
@qsm-odoo Can you check and merge this PR please?

theme_vehicle/static/src/scss/primary_variables.scss
```diff
@@ -117,12 +117,9 @@ $o-website-values-palettes: (
         'btn-padding-x': 1.15rem,
         'btn-border-radius': 1px,
         'btn-border-radius-sm': 0,
-        'btn-border-radius-lg': 2px,
+        'btn-border-radius-lg': 1px,
         'btn-padding-y': .474rem,
-        'btn-padding-y-lg': 1rem,
-        'btn-padding-x-lg': 3.3rem,
         'btn-font-size': 1.2rem,
-        'btn-font-size-lg': 2rem,
```
qsm-odoo, line 125: Hello sorry @xlu-odoo but I tested and with this, the large buttons are really pretty much the same as the normal one... you can reduce the size if needed* but make it so the option still make some sense (= large is bigger than medium).

\* normally we should not merge those anymore as this will impact website currently using the current style...

qsm-odoo (changes requested):

qsm-odoo (approved):
- Changed the commit message to respect the 72 characters limit
- Rebased and retargeted the PR for 15.0 instead of deprecated saas-14.5

@robodoo r+

PR #502, opened 2021-09-24
[FIX] theme_*: move svg preview from iap to design-themes repo

JKE-be:


qsm-odoo (approved):
@robodoo r+

JKE-be:
@robodoo r+

PR #506, opened 2021-09-27
[FIX] *: remove yiq threshold overrides

qsm-odoo:
Many themes were overriding the yiq threshold, thus having an impact
on which of the dark or light text is used over a particuler background
by default. This is a tricky value to change and only made some sense
when the Odoo themes were defining their own color palettes. Now that
all palettes can be used in all themes, it makes no sense to change the
yiq threshold as it induces inconsistent behavior for a same palette
used in different themes. Besides, leaving the value by default is most
of the time what you want and when it is not entirely correct, the user
has full control over the color combination text colors anyway.

task-2655730


qsm-odoo (commented):
@robodoo r+

PR #511, opened 2021-10-04
[FIX] theme_nano: remove breaking container padding removals

qsm-odoo:
Removing the container padding in some cases may be fine but cannot
be done because would create horizontal scrollbars in full-width mode
or simply on mobile devices.

To further improve with task-2241779


qsm-odoo (commented):
@robodoo r+

PR #514, opened 2021-10-05
[REF] theme_*: inject theme's preferred palette into ui

qsm-odoo:
This commit injects for each theme a "preferred" palette into the
website's UI palette list ("Theme" tab).
It also removes '$o-palette-priority-prefix' overrides, letting website
handling the palettes order.

task-2656337

X-original-commit: 0539787e23b02d1ac68eeb9afeb63ecbb3c1c112


qsm-odoo:
@robodoo r+

PR #525, opened 2021-10-27
[IMP] theme_odoo_experts: improve the theme with device shapes

Brieuc-brd:
This commit improves the global layout of **theme_odoo_experts** to give more variety with the others corporate themes (buzzy, graphene, clean,etc.).

The purpose is to have a theme that can be used to present an application or screenshots of a web page and to promote the new category of device shapes from the PR [#79068](https://github.com/odoo/odoo/pull/79068).

task-2666049


qsm-odoo (changes requested):
Cool result :+1: However all the images use a device shape have an undefined "width" when checking the editor panel, could you check? Also some the image of the second snippet could be higher quality I think, it's pixelated even without perspective, don't you think?

Brieuc-brd:
> Cool result 👍 However all the images use a device shape have an undefined "width" when checking the editor panel, could you check? Also some the image of the second snippet could be higher quality I think, it's pixelated even without perspective, don't you think?

Hello @qsm-odoo,
The image is now updated.
About the width issue, I think it's related to the **data-forced-size** attribute introduced in the Community PR, what do you think @detrouxdev ?

Brieuc-brd:
Hello @qsm-odoo,
The issue with the undefined width should be fixed. 👍

qsm-odoo (approved):
@robodoo r+

PR #526, opened 2021-11-04
[FIX] *: increase text contrast for cover snippet

qsm-odoo:
\*: theme_kea, theme_orchid, theme_nano

The theme customization of the cover snippet was not perfect for those
themes. In particular for the Kea theme where the text was not forced to
white anymore (leaving the default black) and the background filter was
pratically removed, leaving the snippet's text properly contrast with
light brackgrounds but be barely visible with dark backgrounds.

Unfortunately, the Kea theme is the most used currently, alongside the
"Software company" industry... which uses a dark background for the
cover image.


rdeodoo (commented):
@robodoo r+

PR #529, opened 2021-12-01
[IMP] theme_kiddo: redesign of the theme

Brieuc-brd:
This commit introduces a redesign of the Kiddo theme.
It also moves 6 svg icons from the "_static/src/img/snippets/_" to "_static/shapes_" to allow the use of the dynamic colors in the Web Editor.

Two new background shapes are introduced with this redesign in #[PR80658](https://github.com/odoo/odoo/pull/80658) in the Community repo 

task-2687297


Brieuc-brd:
@stefanorigano

theme_kiddo/static/src/scss/primary_variables.scss
```diff
@@ -140,13 +140,10 @@ $o-website-values-palettes: (
         'buttons-font': 'Varela Round',
         'link-underline': 'never',
         'header-template': 'default',
-        'footer-template': 'descriptive',
-        'btn-primary-outline': true,
+        'footer-template': 'centered',
     ),
 );
 
-$o-selected-color-palettes-names: append($o-selected-color-palettes-names, 'kiddo-3');
```
stefanorigano, line 148: Is this not necessary anymore?

stefanorigano (commented):
Thanks,

the design is good and I don't see nothing "dangerous" in the code.
Some doubts about how to "migrate" prev. users to this new design,  but I'll let this part to the website team.

The only remark I have is about is about the "bouncing" shapes in the 3 col snippet.
The effect looks nice when applied to a single  image (eg. image-text), but I'm not sure when you have 3 images in a row that bounces in different direction, it's a bit overkill IMO. What do you think?

image

theme_kiddo/static/src/scss/primary_variables.scss
Brieuc-brd: The palette is duplicated when we use this line with a default palette.

Brieuc-brd:
Hey @stefanorigano !
You're right, I replaced them by some static shapes. What do you think ?
Capture d’écran 2021-12-06 à 09 24 55

stefanorigano:
> Hey @stefanorigano ! You're right, I replaced them by some static shapes. What do you think ? Capture d’écran 2021-12-06 à 09 24 55

Looks better to me :)

Brieuc-brd:
Hello @odoo/rd-website,
Is it possible to review this PR ?
Thanks in advance ! 👍

theme_kiddo/static/src/scss/primary_variables.scss
```diff
@@ -184,6 +181,5 @@ $o-gray-color-palettes-compatibility-indexes: (
 // Shapes
 //------------------------------------------------------------------------------
 
-$o-bg-shapes: change-shape-colors-mapping('web_editor', 'Origins/14_001', (3: 2, 4: rgba(0, 0, 0, 0)));
```
qsm-odoo, line 187: Can't remove this, unnecessary break on migration (you can add a comment to say it is kept for compatibility maybe and we'll somehow deal with them in ten years :shrug: ).

theme_kiddo/views/images_library.xml
```diff
@@ -107,36 +142,10 @@
...
-</record>
-<record id="s_product_list_default_image_5" model="theme.ir.attachment">
-    <field name="key">website.s_product_list_default_image_5</field>
-    <field name="name">website.s_product_list_default_image_5</field>
-    <field name="url">/theme_kiddo/static/src/img/snippets/s_product_list_5.svg</field>
-</record>
-<record id="s_product_list_default_image_6" model="theme.ir.attachment">
-    <field name="key">website.s_product_list_default_image_6</field>
-    <field name="name">website.s_product_list_default_image_6</field>
-    <field name="url">/theme_kiddo/static/src/img/snippets/s_product_list_6.svg</field>
```
qsm-odoo, line 139: As always we say it is ok to change all the images of the themes despite the fact that people who kept the themes ones will see their website getting new images at some places (I personally find that not perfect but still ok as people are likely to change the default images and mainly: the problem is now only when they did not use the configurator).

However for these 6 specific SVG ones, the problem is the same with configurator or not and those are images that are more likely to be kept by the user. After migration, the user gets this result (at best):
image

Maybe for these 6 ones we should keep the ir.attachment records and you would just need to update the URLs (and add a comment to say it is for compatibility)?

qsm-odoo (changes requested):
Seems nice, and I did not spot issues on customization xpaths (although I did not spend 5minutes on each line) :+1: 

Just a few compatibility details.

theme_kiddo/views/snippets/s_banner.xml
```diff
@@ -4,34 +4,31 @@
 <template id="s_banner" inherit_id="website.s_banner">
     <!-- Section -->
     <xpath expr="//section" position="attributes">
-        <attribute name="class" add="o_background_video s_parallax_no_overflow_hidden oe_img_bg o_bg_img_center" remove="parallax s_parallax_is_fixed" separator=" "/>
-        <attribute name="data-bg-video-src">//player.vimeo.com/video/60112275?autoplay=1&amp;muted=1&amp;loop=1</attribute>
-        <attribute name="data-oe-shape-data">{"shape":"web_editor/Origins/17"}</attribute>
-        <attribute name="style">position: relative; background-image: url("/web/image/website.s_banner_default_image");</attribute>
+        <attribute name="class" add="s_parallax_no_overflow_hidden o_full_screen_height oe_img_bg o_bg_img_center pb216 pt72 o_cc o_cc5" remove="parallax s_parallax_is_fixed pt96 pb96" separator=" "/>
+        <attribute name="data-scroll-background-ratio">0</attribute>
```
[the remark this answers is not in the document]
Brieuc-brd: By default, the value of data-scroll-background-ratio is « 1 ». If I remove it, the banner won’t have a fixed background but the select box will show « Fixed » in the Web Editor.

qsm-odoo (approved):
@robodoo r+ :+1:

PR #539, opened 2021-12-10
[FIX] theme_*: fix the activation and deactivation of assets

bvr-odoo:
Before this commit:

- The assets were not enabled or disabled when installing themes because
we searched for assets with the "key" field instead of the "name" field.

- The "Ripple Effect" was not correctly installed on several themes.
Because the variable and the assets were not both activated. Even if the
variable is not used, it is still necessary for the proper functioning
of the option (e.g. computeWidgetState).

task-2686370


qsm-odoo (approved):
@robodoo r+

PR #544, opened 2021-12-13
[FIX] theme_*: use correct header SCSS variable

xlu-odoo:
This commit fixes the "Contact" header and the "default" header that were not displaying correctly (in related themes) in the web editor panel as expected.

task-2713578

### DIFF
--- a/theme_anelusia/static/src/scss/primary_variables.scss
+++ b/theme_anelusia/static/src/scss/primary_variables.scss
@@ -8,7 +8,7 @@ $o-website-values-palettes: (
 
         // Header
         'header-font-size':                 1rem,
-        'header-template':                  'template_header_default',
+        'header-template':                  'default',
 
         // Font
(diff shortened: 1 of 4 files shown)


qsm-odoo:
Looks like a fix that should be done in stable if it is right :)

qsm-odoo:
@xlu-odoo Like I said it should be done in stable... like you changed, it's me who apparently forgot the PR afterwards (but ping me when you do the changes I ask :+1:). With the `Contact` header enabled but without the related CSS rules, the off-canvas is broken for example, so this really has to target 15.0. However, you can still rebase (on 15.0) and at the same time: fix the problem everywhere, not only theme_graphene. I see the problem is in multiple themes.

Thanks :+1:

xlu-odoo:
@qsm-odoo Rebased and checked all themes. Found a few others SCSS variable errors. 👍

PR #551, opened 2022-02-23
[FIX] theme_*: Increase page header line height

nboulif:
Steps to reproduce:

  - Install `website_appointment` module
  - Change website theme and select : Cobalt Theme
  - Go to website
  - In menu, click on `Appointment`
  - Select anyone, select any date and confirm appointment

Issue:

  Title text on each other.

Cause:

  The custom page_header class (in portal module) override the default
  line-height with 2.1 rem (instead of 1.1).
  In some themes, the headings tags font-size has been changed with
  3.875rem (instead of 2.5rem).
  => font-size is too big regarding the line-height

Solution:

  Increase line-height in page_header class for needed themes.

opw-2745485


nboulif:
Note: If fix okay, I will update all needed themes.

nboulif:
@rdeodoo Hello :slightly_smiling_face: May I ask for a review please? Thanks in advance!
Note: I'n not used to use SCSS, is there another way to override the `%o-page-header` variable?

theme_cobalt/static/src/scss/portal.scss
```diff
@@ -0,0 +1,8 @@
+//------------------------------------------------------------------------------
+// Fonts
+//------------------------------------------------------------------------------
+
+// Bootstrap hacks
+.o_page_header {
+    line-height: 4.2rem;
+}
```
qsm-odoo, line 8: @nboulif This is not the right fix for multiple reasons:
- Users won't receive the fix as it needs an update of the theme since you create a new asset
- The fix is not specific to a specific theme, as shown in the task's screenshot: as soon as titles are too big, the base CSS rule breaks the layout with an hardcoded line-height. So my guess is that even without theme, if I configure a big font-size there will be the same bug (and even if not, it could be in any custom theme and probably even not only in cobalt in our own themes).
- Even if this was the right way to go (fix specific to Cobalt), you are creating an extra CSS rule to update the base rule. You should not: a variable in the base file and updated in the primary_variables.scss file of cobalt would be the way to go so that no extra CSS rule is created. Themes should not have extra CSS rules anymore, if a CSS rule makes sense it should be in standard parametrized/enabled/disabled by the theme.

Also, what would be the point of commenting the CSS rule with "Bootstrap hacks" while it is a class of ours (not bootstrap) that you just override?

Anyways, in this case I think the fix should simply be in standard, removing the line-height entirely (or at minimum it should not be expressed with a hardcoded unit like your commit message actually says...). Even in website without theme, having a h1 with the o_page_header class leads to broken alignments as shown by this screenshot (the text goes inside the padding/margin):
image

Also after blaming, trying to find some info about the origin of that line-height rule which I was surprised that it could have been part of BS3, I found that it was actually introduced by this commit: https://github.com/odoo/odoo/commit/86ab1c55b4229afb74f630b5041a8354c705f849 -> it could be a useful info to put in the commit message and it is also actually good to find as you could have asked info to SRI if you really were not sure why the rule was there. Apparently I merged it, it was not good: consider removing it after some testing of where it is actually used throughout the Odoo code.

Thanks :+1:

qsm-odoo (changes requested):

PR #554, opened 2022-03-28
[IMP] *: remove _render from ir.ui.view

mart-e:
Follow adaptations made in the render api at odoo/odoo#85110


mart-e:
@robodoo r+

PR #560, opened 2022-04-21
[FIX] website_animate: fix animation in mega menu

bvr-odoo:
Before this commit, the animations were never launched in a mega menu.

Indeed the animations are launched when the element to be animated
appears in the viewport according to the scroll. But in the case of a
mega menu we just want the animation to start when the mega menu is
opened. For that we don't need all the code that checks the scroll, etc.
The animation starts by itself when the element inside the mega menu is
made visible in CSS.

This commit also removes the "Each time it becomes visible" option for
an element animated in a mega menu because in this case this option does
not make sense.

opw-2764895

### DIFF
--- a/website_animate/static/src/js/o_animate.frontend.js
+++ b/website_animate/static/src/js/o_animate.frontend.js
@@ -13,7 +13,7 @@ var WebsiteAnimate = {
     // Retrieve animable elements and attach handlers.
     start: function () {
         var self   = this;
-        self.items = $(".o_animate");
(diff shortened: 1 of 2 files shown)


qsm-odoo:
As the suggested fix here is not fully stable, let's just choose to ignore that problem in 13.0. The version is deprecated in two months and this can be considered a limitation anyway. I'll see to merge the [14.0 PR](https://github.com/odoo/design-themes/pull/561) (and the 15.0 PR has already been merged).

PR #561, opened 2022-04-21
[FIX] website_animate: fix animation in mega menu

bvr-odoo:
Before this commit, the animations were never launched in a mega menu.

Indeed the animations are launched when the element to be animated
appears in the viewport according to the scroll. But in the case of a
mega menu we just want the animation to start when the mega menu is
opened. For that we don't need all the code that checks the scroll, etc.
The animation starts by itself when the element inside the mega menu is
made visible in CSS.

This commit also removes the "Each time it becomes visible" option for
an element animated in a mega menu because in this case this option does
not make sense.

opw-2764895

### DIFF
--- a/website_animate/static/src/scss/o_animate_frontend.scss
+++ b/website_animate/static/src/scss/o_animate_frontend.scss
@@ -11,7 +11,7 @@
     text-rendering: geometricPrecision; // take care of animated titles
     visibility: hidden;
 
-    &:not(.o_animating) {
+    &:not(.o_animating):not(.o_animate_in_dropdown) {
         transform: none !important;
     }
 }
--- a/website_animate/static/src/js/o_animate.editor.js
+++ b/website_animate/static/src/js/o_animate.editor.js
@@ -43,5 +43,22 @@ sOptions.registry.o_animate = sOptions.Class.extend({
             this.$target.toggleClass('o_animate_preview o_animate', !!widgetValue);
         }
     },
+
+    //--------------------------------------------------------------------------
+    // Private
+    //--------------------------------------------------------------------------
+
+    /**
+     * @override
+     */
+    async _computeWidgetVisibility(widgetName, params) {
+        const classes = params.optionsPossibleValues.selectClass;
+        if (classes && classes.includes('o_animate_both_scroll')) {
+            // Hide the "Each time it becomes visible" option if the target is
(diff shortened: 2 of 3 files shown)


website_animate/static/src/scss/o_animate_frontend.scss
```diff
@@ -10,11 +10,12 @@
     backface-visibility: hidden; // avoid flickering
     text-rendering: geometricPrecision; // take care of animated titles
     visibility: hidden;
+}
 
-    &:not(.o_animating) {
```
bvr-odoo, line 14: I first tried to do this here but it doesn't work, I don't know why 
`&:not(.o_animating):not(.dropdown-menu .o_animate)`

qsm-odoo: Actually you are doing that in your final PR code? That cannot work :grimacing: Maybe not ready for review yet?

qsm-odoo: Mmh actually it now works: https://caniuse.com/css-not-sel-list
Even if it matches the browsers we support, it is still very new though. And I still consider it a bad practice if it can be avoided, is there no other way?

bvr-odoo: > Actually you are doing that in your final PR code? That cannot work grimacing Maybe not ready for review yet?

Actually my "try" in the code line was using the "&" (to avoid adding code lines) and not multiple component in a not ;)

bvr-odoo: > Mmh actually it now works: https://caniuse.com/css-not-sel-list Even if it matches the browsers we support, it is still very new though. And I still consider it a bad practice if it can be avoided, is there no other way?

yes for sure, I'll ping you when it's done.
I did this only to reduce the number of lines of code.

bvr-odoo: @qsm-odoo Finally, I can't find a solution in CSS. Do you think it's worth doing this in JS?

qsm-odoo: Maybe it will be even more readable in JS indeed (like adding a special class on animated elements in a dropdown and then adapting the css rules accordingly).

qsm-odoo (approved):
@fw-bot ignore
@robodoo r+

PR #562, opened 2022-05-10
[IMP] test_themes: add a test for the `_post_copy()` community fix

rdeodoo:
https://github.com/odoo/odoo/pull/91012

Now, `_post_copy()` is only called when the module is installed for the
first time on a website, not when the module is updated.

See counterpart commit in community for more details.

opw-2824045

### DIFF
--- a/test_themes/tests/__init__.py
+++ b/test_themes/tests/__init__.py
@@ -2,3 +2,4 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from . import test_crawl
+from . import test_theme_upgrade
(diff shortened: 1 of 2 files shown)


rdeodoo:
@d-fence This is a new standalone test, not yet ready to be merged but that will most likely be soon.

If I am not wrong, we still don't have the `standalone` tests running on the testing suite of at least the staging builds, as discussed in private discord with @Xavier-Do following https://github.com/odoo/odoo/pull/81181/files#r772383106
(I think XDO prefered to have it on every build? I don't really mind either way, just not only in the nightly so we don't allow people to completely break the website app to only figure later when looking at the nightly when it is too late)

If that's not yet done, could you please do it asap?

Thanks

qsm-odoo (approved):
@robodoo r+

PR #563, opened 2022-05-25
[FIX] theme_*: remove some snippets hard coded background color

sobo-odoo:
This is the direct following of the PR [1].

Some themes "hard code" the background colors of some snippets, which
prevents the user to change it with a preset color without deleting the
current background first.

This commit does two things:
- remove the problematic hard coded background colors;
- replace the colors that are removed by some themes by the ones that
were set in [1] (for the searchbar and text highlight snippets).

[1]: https://github.com/odoo/odoo/pull/90749

task-2824393

### DIFF
--- a/theme_bistro/views/snippets/s_image_text.xml
+++ b/theme_bistro/views/snippets/s_image_text.xml
@@ -4,7 +4,7 @@
 <template id="s_image_text" inherit_id="website.s_image_text">
     <!-- Section -->
     <xpath expr="//section" position="attributes">
-        <attribute name="class" add="pt64 pb64 bg-200" remove="pt32 pb32" separator=" "/>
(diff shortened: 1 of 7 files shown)


qsm-odoo (approved):
I push-forced to make this diff https://github.com/odoo/design-themes/compare/9adff2c8921eca30f9a58a6b1d67eb6614ab62ed..728c60a5a6f08b192a26151907cf12e787d80174

(note I left the mention to the community PR even though it would be added automatically too. But as in this case you mention a bit more things, it's fine :+1:)

Note: the design for auto generated homepages via the configurator for theme_treehouse will be a bit less perfect after this IMO (white background and white cards for text image snippet at the end) but I judged the gain in UX better :+1: (cc @Cocographique)

@robodoo r+

PR #572, opened 2022-06-10
[IMP] test_themes: adapt to website_preview client action 

younn-o:
This commit removes the tooltips shown when hovering over a theme in
the website switcher dropdown. This feature will be reimplemented later,
patching the WebsiteSwitcherSystray component.

It also adapts the ?fw= url parameter to force a website id during the
tours.

task-2687506

### DIFF
--- a/test_themes/static/src/js/navbar.js
+++ b/test_themes/static/src/js/navbar.js
@@ -1,30 +1,33 @@
-odoo.define('test_themes.website_selector', function (require) {
-'use strict';
+// TODO: This will be reimplemented following the frontend UI being moved to
+// the backend
 
-const websiteNavbarData = require('website.navbar');
-const { registry } = require("@web/core/registry");
+// odoo.define('test_themes.website_selector', function (require) {
+// 'use strict';
 
-const TestThemesWebsiteSelector = websiteNavbarData.WebsiteNavbarActionWidget.extend({
-    /**
-     * @override
-     */
-    start: function () {
-        this.$('.js_multi_website_switch[data-toggle]').tooltip({
-            html: true,
-            placement: 'left',
-            delay: {show: 100, hide: 0},
-        });
-        return this._super(...arguments);
-    },
+// const websiteNavbarData = require('website.navbar');
+// const { registry } = require("@web/core/registry");
 
-});
+// const TestThemesWebsiteSelector = websiteNavbarData.WebsiteNavbarActionWidget.extend({
+//     /**
+//      * @override
+//      */
+//     start: function () {
+//         this.$('.js_multi_website_switch[data-toggle]').tooltip({
+//             html: true,
+//             placement: 'left',
+//             delay: {show: 100, hide: 0},
+//         });
+//         return this._super(...arguments);
+//     },
 
-registry.category("website_navbar_widgets").add("TestThemesWebsiteSelector", {
-    Widget: TestThemesWebsiteSelector,
-    selector: '#website_switcher',
-});
+// });
 
-return {
-    TestThemesWebsiteSelector: TestThemesWebsiteSelector,
-};
-});
+// registry.category("website_navbar_widgets").add("TestThemesWebsiteSelector", {
+//     Widget: TestThemesWebsiteSelector,
+//     selector: '#website_switcher',
+// });
+
+// return {
+//     TestThemesWebsiteSelector: TestThemesWebsiteSelector,
+// };
+// });
--- a/test_themes/tests/test_crawl.py
+++ b/test_themes/tests/test_crawl.py
@@ -51,4 +51,4 @@ class Crawler(HttpCase):
         Website = self.env['website']
         websites_themes = Website.get_test_themes_websites()
         for website in websites_themes:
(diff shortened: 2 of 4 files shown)


rdeodoo:
Is there a technical issue which prevent it to be converted now?

younn-o:
No, it will be possible but the way we get the preview images changed, and we need to have a DropdownItemWithTooltip component extending the original DropdownItem.
I started to do it but did not finish yet and went back to the community pr comments

test_themes/__manifest__.py
```diff
@@ -41,16 +41,8 @@
         'theme_yes',
         'theme_zap',
     ],
-    'data': [
-        'views/website_navbar_templates.xml',
-    ],
```
qsm-odoo, line 46: Commit message should not mention "website" in apps ? :thinking:

test_themes/static/src/js/navbar.js
```diff
@@ -1,30 +0,0 @@
...
-
-registry.category("website_navbar_widgets").add("TestThemesWebsiteSelector", {
-    Widget: TestThemesWebsiteSelector,
-    selector: '#website_switcher',
-});
-
-return {
-    TestThemesWebsiteSelector: TestThemesWebsiteSelector,
-};
-});
```
qsm-odoo, line 30: I don't mind this being reimplemented post merge but maybe we could preserve some of the history by at least not removing file that will be reintroduced later ? :/

At least this JS file, I would:
- Have a commit which, in this case, just comments the previous code, with a sentence at the top stating "TODO this will be reimplemented following the frontend UI being moved to the backend". That way we never have a point in the history with the feature being gone without any explication.

Post merge:
- Have a commit which only *moves* the file to the right location to be added in the backend assets (-> preserve history, technically, I would have liked that for all website files that were dropped in the community PR too but too late I guess :wink:)
- Have a commit which uncomments and adapts the code.

test_themes/tests/test_crawl.py
```diff
@@ -51,4 +51,4 @@ def test_02_homepage_tour_every_theme(self):
         Website = self.env['website']
         websites_themes = Website.get_test_themes_websites()
         for website in websites_themes:
-            self.start_tour(f"/?fw={website.id}", 'homepage', login='admin')
+            self.start_tour(f"/web?fw={website.id}", 'homepage', login='admin')
```
qsm-odoo, line 54: @rdeodoo I always forget: did not we drop the fw param? :grimacing:

qsm-odoo (changes requested):

test_themes/tests/test_crawl.py
rdeodoo: We did, it got deprecated (not used anymore but kept for stable) with https://github.com/odoo/odoo/commit/2c8c079c059f38b808c73bf17450a2efac2d388b and removed somewhere between 15.0 and master (propably with httpocalypse)

rdeodoo: But the `design-themes` `test_themes` module is adding the support for `fw` so it is fine I guess https://github.com/odoo/design-themes/blob/master/test_themes/models/ir_http.py#L16-L17

qsm-odoo: Although, strange it would work on `/web` with an override to `_frontend_pre_dispatch(cls)` no @younn-o ?

younn-o: I changed it to _pre_dispatch for it to work

rdeodoo: That's a bit concerning.. It means it actually didn't work as expected and didn't change website, but tested the tour on the same default website in each loop?

Generally for other tests of that kind in `test_themes`, I wrote some step/check to ensure we landed on the expected website, eg
```py
self.assertTrue("/%s/static/src" % website.theme_id.name in res.text, "Ensure correct website was loaded")
```

That seems kinda important to fix/ensure here too, if what I said is correct.

Maybe the tour `homepage` (from website module) could be started with a URL query parameter passing the website to do the same check that I mentioned in JS? A new check step in the `homepage` tour checking that the asset URL contain the requested website.

A bit like on the `snippets_all_drag_and_drop` which check something from the URL parameter sent from the python.

Or any other idea, while it ensures we landed on the correct website

rdeodoo: About my previous remark, you can ignore it.
If that's true, that's a concern about stable/master too, the test is not robust enough and won't be red if the `fw` doesn't work anymore (meaning we actually don't really test anything).
I will have a look where the test was introduced, if that's indeed the case.

younn-o: I did a manual check (log the logo[aria-label] during the tour) and the websites are correctly changed during `test_02_homepage_tour_every_theme`
We could also maybe check on the html[data-website-id] ?
And indeed we should use the get_client_action_url method and not /web when starting the tours from python... I checked and for now, our runbot builds are a bit slower than master: 44 => 46 minutes for community and 62 => 66 minutes for enterprise, maybe we can improve with that :crossed_fingers:

qsm-odoo (approved):
@robodoo r+

PR #573, opened 2022-06-29
[IMP] test_themes: (re)add preview to website (themes) switcher

detrouxdev:
Commit [1] moved the website switcher from the frontend to the backend
and in doing so re-factored it into an OWL component.

Initially, this refactoring broke the theme preview from the previous
widget. This commit re-introduces those previews.

[1]: https://github.com/odoo/design-themes/commit/163a8b6fae0d6709d61af81550d1ad77dfe311c9

task-2687506

Co-authored-by: Younn Olivier <yol@odoo.com>


detrouxdev:
@rdeodoo Not that much of a priority but I think I'd be neat to have it back 😄

rdeodoo (changes requested):

rdeodoo (approved):
@robodoo r+ thank :+1: 

I applied the changed myself here https://github.com/odoo/design-themes/compare/b3bc433d2b774330b246e87dba03cbde38a8714c..417cf6dfad8513a251cb38d5e1c2173c5670146e

Note that the comment left in the code during the FE>BE merge was to keep the history. Thus a move commit could have been a good thing but we actually really don't care bout those files which are fully rewritten anyway.

The JS part could probably be improved a bit now that we return a simpler dict but this is LGTM

rdeodoo:
@robodoo r+

PR #581, opened 2022-08-30
[FIX] test_themes: fix theme_upgrade standalone test

rdeodoo:
Ripple option is now an ir.asset, not an ir.ui.view anymore, see [1].

[1]: https://github.com/odoo/odoo/commit/9f56357cc1f4a7b8606ef4d5fd431fc396bdf1e8

### DIFF
--- a/test_themes/tests/test_theme_upgrade.py
+++ b/test_themes/tests/test_theme_upgrade.py
@@ -25,13 +25,13 @@ def test_01_theme_upgrade_post_copy(env):
         except Exception:
             pass
 
-    ripple_specific_view = Website.viewref('website.option_ripple_effect')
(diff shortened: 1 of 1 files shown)


rdeodoo:
@robodoo r+

cc @qsm-odoo indeed one day it would be more robust to make the test rely on a test theme instead of a real theme that might change in the future.
Good news, I have a new (test) theme coming in another PR where I create everything possible and test it (mega menu, submenu, new menu hierarchy etc).

rdeodoo:
cc @odoo/rd-runbot this should be enough to fix the nightly website standalone build
I am pretty sure the error was something else than what the nightly shows now but couldn't find a very old build to confirm
This PR is making the tour working locally and the website standalone build succeed on my branch :crossed_fingers:

PR #582, opened 2022-08-30
[FIX] test_themes: force request's website before _frontend_pre_dispatch

rdeodoo:
The callstack is:
- _pre_dispatch() (http_routing)
  - super()
  - _frontend_pre_dispatch()
    - super()
      - _frontend_pre_dispatch() (website) -> Set the website on request

But in `test_themes` we were forcing the website in a `_pre_dispatch()`
override, after calling `super()`.
It means that the call to `_frontend_pre_dispatch()` was actually done
without having yet forced the website in session. get_current_website()
would then not consider the `fw` param as not yet in session, and the
request.website would be set to 1, the default one.
Later in the business code, there would be a mismatch between
`request.website` and `get_current_website()`.

### DIFF
--- a/test_themes/models/ir_http.py
+++ b/test_themes/models/ir_http.py
@@ -10,9 +10,9 @@ class Http(models.AbstractModel):
 
     @classmethod
     def _pre_dispatch(cls, rule, args):
-        super()._pre_dispatch(rule, args)
-
         # Allow public user to use `fw` query string in test mode to ease tests
         force_website_id = request.httprequest.args.get('fw')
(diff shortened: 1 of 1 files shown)


rdeodoo:
@qsm-odoo You'd need to r+ this one and the enterprise one for the whole bundle to be merged (upgrade and community are still in r+ mode :+1: )

qsm-odoo (approved):
@rdeodoo This is only a master bug?
@robodoo r+

qsm-odoo:
@robodoo r+ please

PR #586, opened 2022-08-31
[IMP] test_themes, theme_test_custo: add new test theme for tests 

rdeodoo:
This commit introduces a new theme for tests purpose.
The goal is to be able to create records and simulate a real use case
without only testing through unit tests in python.
For instance, with this commit, we now fully test:
- menu creation.
- page creation.

This will be extended later to be able to test other stuff too, maybe
like changing the website logo (not supported yet). Also, our existing
standalone test in themes should use this theme instead of nano to not
rely on a real theme that might change.

Courtesy of NMI as this theme is based on his job.


rdeodoo:
This took way, way, way more time that expected / should have.

Anyway, this should be good now, and we now have a fake theme where we can insert our theme specific test logic :hot_face: 

I'll make sure the runbot is green and re-read everything before this can be properly reviewed.

test_themes/__init__.py
```diff
@@ -10,11 +10,8 @@ def post_init_hook(cr, registry):
     ''' Create a new website for each theme and install the theme on it. '''
     env = api.Environment(cr, SUPERUSER_ID, {})
     IrModule = env['ir.module.module']
-
-    themes = IrModule.search([
-        ('category_id', 'child_of', env.ref('base.module_category_theme').id),
-    ], order='name')
-    exclude_list = ['_common', '_blog', '_sale']
+    themes = IrModule.search(IrModule.get_themes_domain(), order='name')
+    exclude_list = ['_common']
```
qsm-odoo, line 14: Mmh, did not you say in the community commit that excluding theme_common is already done by `get_themes_domain`?

theme_test_custo/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+{
+    'name': 'Custom Theme (Testing suite)',
+    'summary': '',
+    'description': '',
+    # Remove the `/Hidden` part to make it selectable for tests purpose
+    'category': 'Theme/Hidden',
+    'version': '15.0.0',
```
qsm-odoo, line 7: Version 15? ^^ I missed the first 14 :o

theme_test_custo/data/images.xml
```diff
@@ -0,0 +1,12 @@
...
+
+<record id="logo" model="theme.ir.attachment">
+    <field name="name">Test Theme Custom Logo</field>
+    <field name="url">/theme_test_custo/static/src/img/logo.jpg</field>
+    <field name="key">theme_test_custo.logo</field>
+</record>
+
+<!-- TODO: It is not yet possible to alter the website logo -->
+
+</odoo>
```
qsm-odoo, line 12: I would remove that, can't be sure we'll allow to do that IMO. Except if that means less code to allow it but I don't think so.

theme_test_custo/static/src/scss/primary_variables.scss
```diff
@@ -0,0 +1,108 @@
...
+        'headings-font':                    'Poppins',
+        'navbar-font':                      'Source Sans Pro',
+        'buttons-font':                     'Source Sans Pro',
+
+        // Header
+        'header-template':                  'custom-theme',
+        'logo-height':                      4rem,
+
+        // Footer
+        'footer-template':                  'custom-theme',
```
qsm-odoo, line 21: I keep repeating not to add 500 spaces to aligns the values in maps, maybe let's not do it here either to prevent this to be copy/pasted elsewhere? :)

theme_test_custo/static/src/scss/primary_variables.scss
```diff
@@ -0,0 +1,108 @@
...
+    'Lobster': (
+        'family': ('Lobster', sans-serif),
+        'url': 'Lobster:400',
+        'properties' : (
+            'base': (
+                'font-size-base':           1rem,
+            ),
+        )
+    ),
+);
```
qsm-odoo, line 73: Oups? Duplicated code, `$o-theme-font-configs` copy/pasted twice :grimacing: You can also drop the "properties" key in both fonts

theme_test_custo/views/header.xml
```diff
@@ -0,0 +1,55 @@
...
+</template>
+
+<!-- Theme Header - Layout -->
+<record id="template_header_custom" model="theme.ir.ui.view">
+    <field name="name">Test Theme Custom Header Template</field>
+    <field name="type">qweb</field>
+    <field name="key">theme_test_custo.template_header_custom</field>
+    <field name="priority">16</field>
+    <field name="inherit_id" ref="website.layout"/>
+    <field name="mode">extension</field>
```
qsm-odoo, line 21: Should we use `<template>` everywhere to encourage using it instead ?

qsm-odoo (changes requested):
Some remarks, did not tested, seems nice :+1:

theme_test_custo/__manifest__.py
rdeodoo: Woopsie :innocent:

theme_test_custo/data/images.xml
rdeodoo: Sure :+1: 
We will be able to if that's part of the "style in db" task, isn't it?

rdeodoo (commented):
Changes done (typo in community too).

theme_test_custo/data/images.xml
qsm-odoo: We could be able to with more code but I don't see the point in having a theme setting up a logo to be honest. If it was 0 line to allow it, sure. But I would not write one line of code to allow it TBH :shrug:

rdeodoo: I don't share the same opinion, as I have always said, the incoming success (or not) of Odoo or any such software is bound to its community themes capability.

I've always been convinced that if wordpress and bootstrap became what they became, its partly due to their template ecosystem, not the other way around.

That's just my deep belief of course.

And I have never seen a wordpress template or a bootstrap template without a custom logo :grimacing: 

AFAIC, this should 100% be possible and it is really important

qsm-odoo (approved):
@robodoo r+

Nice job ! :)

theme_test_custo/data/images.xml
qsm-odoo: I don't see how the marketing of a theme is related to its use. You know I share your opinion about themes, having a logo of the theme set as default for your website has nothing to do with having a community that build themes. They are free to use a custom logo in their demo/marketing pages.

In the end, dumb debate, a community theme has already the possibility to use a custom header, they can add a custom editable image with a custom logo in there if they want (just like we do in our footers instead of using the website logo field). The website logo field will be removed at some point anyway I am sure (I am not sure it has any use case anymore actually? Except showing up in the website form view nobody ever see maybe?).

qsm-odoo: EDIT 2: mmh ok, it has some uses as repeated between the navbar and the off canvas menu, also would create a cow'ed view if edited as a non-field image and probably would be more annoying for the configurator automatically setting up the logo

qsm-odoo: Canceled two years ago: https://www.odoo.com/web#id=2212423&menu_id=4720&cids=1&action=4043&model=project.task&view_type=form but that can change :shrug:

theme_test_custo/__manifest__.py
```diff
@@ -0,0 +1,24 @@
+{
+    'name': 'Custom Theme (Testing suite)',
+    'summary': '',
+    'description': '',
+    # Remove the `/Hidden` part to make it selectable for tests purpose
```
rdeodoo, line 5: Note that for runbot, you obviously can't change this, so one way is to:
- Go to the `act_window` `website.theme_install_kanban_action` (`Pick a theme`)
- Change the `domain` field to keep only the `[('category_id.parent_id', '=', xx)]` part

Then you should be able to select that theme in the theme selection screen

PR #587, opened 2022-09-09
[FIX] theme_test_custo: add bigger timeout to the new test

rdeodoo:
New test was introduced with [1] but it actually sometimes lead to timeout errors.
That's a common issue since we merged the frontend in the backend and some tour (like this one) now have to start both the backend and the frontend (in the iframe).

10 seconds is not enough to load both, it often miss just half a second.

[1]: https://github.com/odoo/design-themes/commit/0e2497b8e84eab8957028f4d63ff0899a9fe6ab3

runbot-4684

### DIFF
--- a/theme_test_custo/static/tests/tours/theme_menu_hierarchies.js
+++ b/theme_test_custo/static/tests/tours/theme_menu_hierarchies.js
@@ -9,6 +9,7 @@ wTourUtils.registerEditionTour('theme_menu_hierarchies', {
     {
         content: 'Check Mega Menu is correctly created',
         trigger: 'iframe #top_menu a.o_mega_menu_toggle',
+        timeout: 20000,
     }, {
         content: 'Check Mega Menu content',
         trigger: 'iframe #top_menu div.o_mega_menu.show .fa-cube',


theme_test_custo/static/tests/tours/theme_menu_hierarchies.js
```diff
@@ -9,6 +9,7 @@ wTourUtils.registerEditionTour('theme_menu_hierarchies', {
     {
         content: 'Check Mega Menu is correctly created',
         trigger: 'iframe #top_menu a.o_mega_menu_toggle',
+        timeout: 20000,
```
qsm-odoo, line 12: Oh indeed, when I first saw the runbot error I though that it was our automatic 30000 timeout of `registerEditionTour` which was still not good enough... but in fact it is only when you set `edition: true` :grimacing: Maybe this should have been done in `registerEditionTour` for when `edition: false` ? Increase the given first step's timeout ?

Anyways, let's r+ this one for now. We can still improve later and of course try to reduce the need for those timeout :grimacing:

qsm-odoo (approved):
@robodoo r+

PR #590, opened 2022-09-21
[IMP] theme_*: add precondition on tour to ensure expected theme

bso-odoo:
This commit adds a precondition to verify the presence of the specific theme at the beginning of their tour.

See https://github.com/odoo/odoo/pull/100690.

task-2687506


rdeodoo (approved):
@robodoo r+

PR #591, opened 2022-09-22
[IMP] theme_*: ease defining only one font family to use in a theme

qsm-odoo:
Before this commit, if a theme wanted to define a main font for their
theme and let the "headings", "navbar" and "buttons" ones use the same
they had to add this in their map:

```
'font': XXX,
'headings-font': null,
'navbar-font': null,
'button-font': null,
```

Indeed, without setting the "null" values, those font would use YYY
which is the first font defined in the theme config, which might be
different from XXX. And if forcing XXX 4 times, like most themes do at
the moment, if the user wanted to change them all, he had to change the
4 ones instead of the main one.

Now, only the 'font' value use the first font defined in the theme
config if not explicitly set. So forcing the null values is not
necessary, it will be the default behavior.
The advantage is thus also functional, as most theme forced all their 4
fonts so changing the main one did not change the others.

### DIFF
--- a/theme_anelusia/static/src/scss/primary_variables.scss
+++ b/theme_anelusia/static/src/scss/primary_variables.scss
@@ -13,8 +13,6 @@ $o-website-values-palettes: (
         // Font
         'font':                             'Source Sans Pro',
         'headings-font':                    'Spartan',
-        'navbar-font':                      'Source Sans Pro',
-        'buttons-font':                     'Source Sans Pro',
 
         // Links
         'link-underline':                   'never',
--- a/theme_artists/static/src/scss/primary_variables.scss
+++ b/theme_artists/static/src/scss/primary_variables.scss
@@ -97,8 +97,6 @@ $o-website-values-palettes: (
         'header-font-size': (13 / 16) * 1rem,
         'font': 'Montserrat',
         'headings-font': 'Oxygen',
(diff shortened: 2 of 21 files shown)


theme_clean/static/src/scss/primary_variables.scss
```diff
@@ -195,9 +195,6 @@ $o-theme-font-configs: (
 $o-website-values-palettes: (
     (
         'color-palettes-name': 'clean-1',
-        'headings-font': 'Dosis',
-        'navbar-font': 'Dosis',
-        'buttons-font': 'Dosis',
```
rdeodoo, line 200: This shouldn't have been removed ?

rdeodoo (changes requested):

theme_clean/static/src/scss/primary_variables.scss
qsm-odoo: Whoa, good catch, [I restored it](https://github.com/odoo/design-themes/compare/3687dea6e6c879300db90244b87bd52a055ba4c4..502d46cc9dd20c14c6e22b8bb4e36fa872c1c8cd) :+1:

rdeodoo (approved):
@robodoo r+

rdeodoo:
Rebuilt it

PR #592, opened 2022-09-23
[IMP] theme_*: adapt some themes to the grid-only Masonry snippet

sobo-odoo:
In [1], the Masonry snippet has been modified to be only in grid mode and its templates have therefore been changed.

This commit adapts the themes that modify the Masonry templates to be in accordance with these changes.

[1]: https://github.com/odoo/odoo/pull/99857

task-2973198


qsm-odoo (approved):
@robodoo r+

PR #595, opened 2022-09-26
[FIX] test_themes: patch website switcher template

bso-odoo:
In [1] when all modules were migrated to the new assets system, the modules in design themes were forgotten.

This commit puts the owl templates into the correct asset.

[1]: https://github.com/odoo/odoo/commit/5410b7c238f211ced2519e53afc99c8468e1f527


rdeodoo (changes requested):
I doubt you are referring to the correct commit, the [commit you mention](https://github.com/odoo/odoo/commit/03641610c2b4a6b832ea87795fd21e1d7af10b59) is a commit of april 2021.

What you want to refer is probably the recent one of @Gorash which removed the qweb assets a few days ago.

It just broke very recently, we fixed it after frontend > backend with https://github.com/odoo/design-themes/commit/fb23f19c1bcf861ae63573d435f6dbdf4b6d6a64

test_themes/__manifest__.py
```diff
@@ -48,7 +48,7 @@
         'web.assets_backend': [
             'test_themes/static/src/js/systray_items/website_switcher.js',
         ],
-        'web.assets_qweb': [
+        'website.assets_editor': [
```
rdeodoo, line 51: Any reason not to have both the xml and js file in the same bundle, like
```py
'web.assets_editor': [
    'test_themes/static/src/js/systray_items/*',
],
```
And maybe you can also remove the `js` folder to fit the community structure ?

rdeodoo (changes requested):

test_themes/__manifest__.py
bso-odoo: @rdeodoo Change done.

rdeodoo: Thank you, it was a bit nitpicking but at least now it is aligned with the rest :)

rdeodoo (approved):
@robodoo r+
Tested with all user rights :+1:

PR #605, opened 2022-10-14
[FIX] theme_bookstore: correct JSON value

can-odoo:
Description of the issue/feature this PR addresses:
---
Broken Pricelist snippet for `theme_bookstore`, as soon as drag and drop

Current behavior before PR:
---
As JSON is invalid, it can not [parse](https://github.com/odoo/odoo/blob/15.0/addons/web_editor/static/src/js/editor/snippets.options.js#L6412) and throws an error `not valid JSON`

Desired behavior after PR is merged:
---
An error occurred due to an incorrect way of passing JSON to the snippet arch, need to correct it in a proper format to achieve the desired result

PR: [605](https://github.com/odoo/design-themes/pull/605)
TaskID: 3002630

### DIFF
--- a/theme_bookstore/views/snippets/s_product_catalog.xml
+++ b/theme_bookstore/views/snippets/s_product_catalog.xml
@@ -5,7 +5,7 @@
     <!-- Section -->
     <xpath expr="//section" position="attributes">
         <attribute name="class" add="o_cc o_cc5 pb112 pt112" remove="pb32 pt48" separator=" "/>
-        <attribute name="data-oe-shape-data">data-oe-shape-data="{"shape":"web_editor/Origins/17","flip":["y"]}"</attribute>
+        <attribute name="data-oe-shape-data">{"shape":"web_editor/Origins/17","flip":["y"]}</attribute>
     </xpath>
     <!-- Shape and filter -->
     <xpath expr="//section/*" position="before">


rdeodoo (approved):
@robodoo r+

Hello, thanks for the fix.

A few things here, that I just applied through a push force:
- You added the `TaskId: 3002630` on the PR message but not in the commit message where it should be and is the most important
- It is `task-xxx` not `TaskId: xxx`, as per our commit guidelines. Otherwise is sometimes create links to commit which hash start with the opw number
- Don't go over 72 char / line in your commit message
- Try to blame and link the commit which introduced the error, in this case https://github.com/odoo/design-themes/commit/1f632d8dd79ceba6e152bedcb6de21cd9ada14c0
- Ideally use markdown to reference commit, eg `[1]:`, see how I adapted it if you are interested
- Your commit was saying it was crashing when "as soon as drag and drop", but it is when clicked/focused after drag and drop?
- Nitpicking: could've written a more specific commit title mentionning the `shape`

In case you are interested:
Before:
```
[FIX] theme_bookstore: correct JSON value
Description of the issue/feature this PR addresses:
---
Broken Pricelist snippet for `theme_bookstore`, as soon as drag and drop

Current behavior before PR:
---
As JSON is invalid, it can not parse and throws an error `not valid JSON`

Desired behavior after PR is merged:
---
An error occurred due to an incorrect way of passing JSON to the snippet arch,
need to correct it in a proper format to achieve the desired result
```

After:
```
[FIX] theme_bookstore: correct shape data JSON value

Since it was introduced with [1], the bookstore `Price` snippet couldn't
be clicked on.
Indeed, a typo/wrong format inserted inside a `data-oe-shape-data` xpath
was making it crash due to invalid JSON format.

[1]: https://github.com/odoo/design-themes/commit/1f632d8dd79ceba6e152bedcb6de21cd9ada14c0

task-3002630
```

can-odoo:
Hello @rdeodoo
> A few things here, that I just applied through a push force:

I will keep those things in mind, and thanks much for guiding me.

PR #616, opened 2022-10-31
[FIX] theme_zap: remove `contentenditable` property from Masonry image

sobo-odoo:
In [1], in order to allow back the edition of the images of the columns in grid mode containing only an image, the `contenteditable` property is removed from these columns.

This commit removes the `contenteditable` property that was set on an image column of the Masonry snippet in the Zap theme.

[1]: https://github.com/odoo/odoo/pull/103767

opw-3028116


qsm-odoo (approved):
@robodoo r+

PR #618, opened 2022-11-24
[REF] theme_common: remove (deprecated) $target in public widgets

qsm-odoo:
See community commit for details.


qsm-odoo (commented):
@robodoo r+

PR #619, opened 2022-12-05
[FIX] theme_kea: fix faulty shape on s_picture

detrouxdev:
Related community PR: https://github.com/odoo/odoo/pull/107224

Commit [1] refreshed the design of the KEA theme but unfortunately introduced the wrong path to a shape on the s_picture snippet.

This causes a traceback when the bg_shape option is toggle on a block inserted directly under that snippet.

This commit fixes the path to the shape.

[1]: https://github.com/odoo/design-themes/commit/88b81d41dadbfaad5d0b36ba2f1c595e309438ba

opw-3082292

### DIFF
--- a/theme_kea/views/snippets/s_picture.xml
+++ b/theme_kea/views/snippets/s_picture.xml
@@ -5,6 +5,7 @@
     <!-- Section -->
     <xpath expr="//section" position="attributes">
         <attribute name="class" add="o_cc3" remove="o_cc2" separator=" "/>
+        <!-- TODO: adapt in master, this was fixed in JS -->
         <attribute name="data-oe-shape-data">{"shape":"web_editor/Origins/Wavy_03","flip":["y"]}</attribute>
     </xpath>
     <!-- Shape -->


theme_kea/views/snippets/s_picture.xml
```diff
@@ -5,7 +5,7 @@
     <!-- Section -->
     <xpath expr="//section" position="attributes">
         <attribute name="class" add="o_cc3" remove="o_cc2" separator=" "/>
-        <attribute name="data-oe-shape-data">{"shape":"web_editor/Origins/Wavy_03","flip":["y"]}</attribute>
+        <attribute name="data-oe-shape-data">{"shape":"web_editor/Wavy/03","flip":["y"]}</attribute>
```
qsm-odoo, line 8: > Is this enough for a fix in stable?

It would be acceptable but not enough IMO:
- People could have done an xpath on this, so changing this could break it. It's however acceptable as unlikely and we normally accept the fact people should not xpath specific things like this.
- Not enough: having to -u a theme for a traceback to stop happening is ok but only if there is no alternative IMO.

In this case, I would avoid both. IMO, we don't care to uglyfy the 16.0 code if the master code is fixed the right way in the forward-port PR. You could do:
- Next to the code calling the `onBuilt` method of options in the website code, check if `data-oe-shape-data="{"shape":"web_editor/Origins/Wavy_03","flip":["y"]}"`, if yes -> fix it.

> This causes a traceback when the bg_shape option is toggle on a block inserted directly under that snippet.
- Fix that with a crappy JS patch too: will avoid a traceback for users having already added a snippet like this in their page.

Some lines of JS code to patch the broken behavior seamlessly for the user in stable and fix it the right way in master is always worth it IMO. (Of course add a comment next to the broken line and the patching lines (something like `TODO adapt in master: this was fixed in JS` / `TODO remove me in master after the XML is patched` / ...)

qsm-odoo (changes requested):

theme_kea/views/snippets/s_picture.xml
detrouxdev: I did something similar to what you suggested but on start + on the initial computation of the templates so that it's only done once at the start of the edition.

Here's the PR: https://github.com/odoo/odoo/pull/107224

qsm-odoo: Don't forget to update the theme one as I said as well ;)

qsm-odoo (approved):
Commit message was only adapted in community and not here... I did it :grimacing: 

@robodoo r+

PR #632, opened 2023-02-02
[FIX] website_animate: fix wrapwrap overflow and animations

bvr-odoo:
Before this commit, on iPhone 8 (and lower) it was possible to scroll
the page to the right when there were animated elements in the page.

A "transform: none" property was applied to non-visible animated
elements to prevent the page from expanding to the right. However, this
property wasn't properly overriding keyframe transforms on iPhone 8 and
lower. This has been resolved by adding "overflow-x: hidden" on the
wrapwrap in case "transform: none" is not applied correctly.

Steps to reproduce the issue:

- On iPhone 8 (Safari).
- Drop a few snippets into a page.
- Add a "Fade In-Right" animation to one of the snippets.
- Scrolls the page so that the animated element is invisible.
- Bug => a horizontal scrollbar appears and it is possible to scroll the
page to the right.

opw-3165651

### DIFF
--- a/website_animate/static/src/js/o_animate.frontend.js
+++ b/website_animate/static/src/js/o_animate.frontend.js
@@ -15,6 +15,13 @@ var WebsiteAnimate = {
         var self   = this;
         self.$scrollingElement = $().getScrollingElement();
         self.items = $("#wrapwrap .o_animate");
+        // Fix for "transform: none" not overriding keyframe transforms on
+        // iPhone 8 and lower.
+        self.forceOverflowXHidden = false;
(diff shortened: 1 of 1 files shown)


bvr-odoo:
:information_source:   PR for V15 => https://github.com/odoo/odoo/pull/111750 (as the animations are no longer in design themes in this version).

qsm-odoo (approved):
@robodoo r+
@fw-bot ignore

(15.0 version here: https://github.com/odoo/odoo/pull/111750 as website_animate was moved in community over there)

Issue #574, opened 2022-06-30
Rename Muli to Mulish?


rdeodoo:
See https://github.com/odoo/design-themes-old/pull/432
Seems to still be needed: https://github.com/odoo/design-themes/blob/15.0/theme_clean/static/src/scss/primary_variables.scss#L153

rafaelbn:
Yes 👍

PR #634, opened 2023-03-07
[IMP] theme_beauty, *: rename the font "Muli" to "Mulish"

rdeodoo:
* theme_bookstore, theme_clean, theme_enark, theme_kiddo, theme_loftspace, theme_odoo_experts, theme_orchid

This commit will rename the font "Muli" to "Mulish" as it has been renamed in Google Fonts.

Closes https://github.com/odoo/design-themes/issues/574

### DIFF
--- a/theme_beauty/static/src/scss/primary_variables.scss
+++ b/theme_beauty/static/src/scss/primary_variables.scss
@@ -42,9 +42,9 @@ $o-theme-font-configs: (
         'family': ('Questrial', sans-serif),
         'url': 'Questrial:300,300i,400,400i,700,700i',
     ),
-    'Muli': (
-        'family': ('Muli', sans-serif),
-        'url': 'Muli:300,300i,400,400i,700,700i',
+    'Mulish': (
+        'family': ('Mulish', sans-serif),
+        'url': 'Mulish:300,300i,400,400i,700,700i',
     ),
     'DM Serif Display': (
         'family': ('DM Serif Display', serif),
--- a/theme_bookstore/static/src/scss/primary_variables.scss
+++ b/theme_bookstore/static/src/scss/primary_variables.scss
@@ -42,9 +42,9 @@ $o-theme-font-configs: (
         'family': ('Questrial', sans-serif),
         'url': 'Questrial:300,300i,400,400i,700,700i',
     ),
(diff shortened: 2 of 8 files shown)


rdeodoo:
Despite not being found in google (both by URL tweak or by search), it is still served for now.

Let's clean that now so we don't forget about it, in case they later remove the `Muli` compatibility.
Probably could be done in earlier version but better than nothing for now.
In 3 years we probably won't even think about it anymore and all stable version will have it :+1:

qsm-odoo (changes requested):
What about doing what JKE said here: https://github.com/odoo/design-themes-old/pull/432#issuecomment-755237524 (using `$o-fonts-similar`)? (alongside what you did here of course). Without that, a customer migrating to 17.0 will have a broken font :wink: (until we can properly migrate them).

And if you do that, I guess it would be nice to just target 16.0 with this. Would allow to have one extra year of safety in case Google drops the support for Muli :shrug: (and I am sure the 16.0 and master version is the same, not worth it checking before).

qsm-odoo:
@rdeodoo 

> Why? It is just a text entry which at the end calls a google font URL with that name?

No, it won't :grimacing: By doing this change:
image

You change the name of the font entry in that map. A user who chose it have this in its user_values.scss file:
```scss
'font': 'Muli',
```
Since you rename the 'Muli' key to something else, the migrated user won't have a font anymore because there won't be any 'Muli' entry in the maps you modified.

Arguably, you could only change the "family" and "url" values and leave the key untouched, but that would lead to displaying the wrong name in the editor.

> We could take more time to write an upgrade script [...]

No, as always: we don't have access to migrate this as this is in attachments.

> [...] and a compatibility stuff in scss indeed but then this will postpone this fix

That's what I am suggesting, there is already something for that: just one entry added in the `$o-fonts-similar` map will solve all issues here :wink: You can r+ without it if you prefer not spending the extra time... but just know that 16.2/3 users will lose their font immediately (not when Google drops his support).

rdeodoo:
> there won't be any 'Muli' entry in the maps you modified.

Indeed, that's what I was missing :+1: 

Created the community PR too

qsm-odoo (approved):
@robodoo r+

PR #640, opened 2023-03-29
[IMP] theme_yes: change description of masonry snippet

can-odoo:
Introduced with [1], same description for Caterers and Venues block.

[1]: https://github.com/odoo/design-themes/commit/299d504df31bf179775eec1d37e71605c3637460

PR: [640](https://github.com/odoo/design-themes/pull/640)
task-2994239

### DIFF
--- a/theme_yes/i18n/theme_yes.pot
+++ b/theme_yes/i18n/theme_yes.pot
@@ -119,7 +119,7 @@ msgstr ""
 
 #. module: theme_yes
 #: model_terms:theme.ir.ui.view,arch:theme_yes.s_masonry_block_default_template
-msgid "Check out our list of favorite venues."
+msgid "Check out our list of favorite caterers."
 msgstr ""
 
 #. module: theme_yes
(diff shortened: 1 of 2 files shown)


rdeodoo (changes requested):
Thanks but you also need to update the .pot file.
As this is dummy data which purpose is to be edited, I'am ok with 15.0

can-odoo:
Hello @rdeodoo, `.pot` file updated

theme_yes/i18n/theme_yes.pot
```diff
@@ -119,7 +119,12 @@ msgstr ""
 
 #. module: theme_yes
 #: model_terms:theme.ir.ui.view,arch:theme_yes.s_masonry_block_default_template
-msgid "Check out our list of favorite venues."
+msgid "Check out our list of favorite caterers."
+msgstr ""
+
+#. module: theme_yes
+#: model_terms:theme.ir.ui.view,arch:theme_yes.s_masonry_block_default_template
+msgid "Find a Caterer"
```
rdeodoo, line 127: Could you please adapt `.pot` so it is the same as the one generated through the UI (settings>translations>export)?
Those should be ordered alphabetically, just generate the .pot from Odoo and check in a text compare tool where should be placed the string you added.
Otherwise, it will keep introducing fake diff whenever someone will generate the .pot for their own modifications.

rdeodoo (changes requested):

theme_yes/i18n/theme_yes.pot
can-odoo: Hello, the `.pot` file was updated based on generated through the UI.

PR #659, opened 2023-05-26
Staging.15.0

MarioAbad:


rdeodoo (changes requested):
Hello,

I think you selected the wrong target (staging.15.0), I changed it to 15.0.

Here are a few unorganized feedbacks:
- We won't introduce new themes in 15.0. Arguably 16.0 or saas 16.X could be considerate candidate, but master is surely the way to go.
- It's unlikely that we will introduce a community theme in our codebase for now. We have been working the past years on the website and themes so the community is able to create easily themes. We have still a few ongoing tasks to facilitate that process for Odoo 17.
So, we are happy to see people working on themes ! But it's too soon and it won't probably fit in our own codebase.
- We don't yet have themes related to the ecommerce directly, especially not with python code. It doesn't really work with our theme selection screen which is showing all themes. In your case, a website only (not ecommerce) database would see your theme and by selecting it, it will most likely install the ecommerce app which is not what they want.
What you want to do is probably create a theme only related to the website, and have a bridge module adding stuff into this theme with a dependency of your theme and the sale/blog/whatever module you want to improve.
- Themes should ideally be only about style and feature-less, which is not the case here as there is many python code.
- I just scrolled quickly but I don't think your theme is multi website friendly/compliant, which is a really important thing regarding themes. I've seen act windows, list views etc, those are not multi website compliant.
You can have a look by creating a second website on a second domain, your theme should only impact the website where you installed it.

Everything I am telling you here is about having a theme in Odoo codebase. It does not apply to the Odoo app stores where your theme would fit :+1: 

Finally, I am not sure if you have seen it, but a theme documentation was released very recently, it might help: https://www.odoo.com/documentation/15.0/developer/howtos/website_themes.html

Anyway, good job, that's a lot of work here :+1: 

cc @qsm-odoo @Cocographique @pierre-pde

PR #661, opened 2023-06-13
[FIX] theme_test_custo: add test for the mobile/desktop theme preview

robinlej:
With the Milk redesign in [1], the possibility to use the mobile/desktop
switcher on a theme preview to see how it behaves/displays on those
devices didn't work anymore.

It has been fixed in [2], but in order to prevent future breaks, this
commit tests this functionality.

task-3355253

[1]: https://github.com/odoo/odoo/commit/caef16ee4e5351142ef04f02798999ccb38a56bc
[2]: https://github.com/odoo/odoo/commit/9445f61bcaee88bc6ac2fc373d7eca4e4b38c154

### DIFF
--- a/theme_test_custo/static/tests/tours/website_theme_preview.js
+++ b/theme_test_custo/static/tests/tours/website_theme_preview.js
@@ -1,0 +1,41 @@
+/** @odoo-module */
+
+import tour from "web_tour.tour";
+
+tour.register("website_theme_preview", {
+    test: true,
+    url: "/web#action=website.action_website_configuration",
+},
+[{
+    content: "Click on create new website",
+    trigger: 'button[name="action_website_create_new"]',
+}, {
+    content: "insert website name",
+    trigger: '[name="name"] input',
+    run: "text Website Test",
+}, {
+    content: "Validate the website creation modal",
+    trigger: "button.btn-primary",
+},
+// Configurator first screen
+{
+    content: "Click Skip and start from scratch",
+    trigger: "button:contains('Skip and start from scratch')",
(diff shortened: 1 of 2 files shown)


theme_test_custo/tests/test_theme_creation.py
```diff
@@ -2,7 +2,7 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from odoo.tests import HttpCase, tagged
-
+from odoo.addons.website.tests.test_configurator import TestConfiguratorCommon
```
rdeodoo, line 5: Next time pay attention to alphabetical order

theme_test_custo/static/tests/tours/website_theme_preview.js
```diff
@@ -0,0 +1,41 @@
...
+    trigger: '[name="name"] input',
+    run: "text Website Test",
+}, {
+    content: "Validate the website creation modal",
+    trigger: "button.btn-primary",
+},
+// Configurator first screen
+{
+    content: "Click Skip and start from scratch",
+    trigger: "button:contains('Skip and start from scratch')",
```
rdeodoo, line 23: This is usually not great to use wording to find elements in tests, we might change the wording or fix a typo later which would break the test.
It's fine tho, but wanted let you know that you could have avoided that by writing the test another way: simply enter edit mode, third tab, click on "Switch theme", which is another (and simpler/shorter) way to access this theme selection screen.

rdeodoo (approved):
@robodoo r+

PR #673, opened 2023-07-26
[IMP] theme_monglia: use BS mixin to have responsive font sizes

Guillaume-gdi:
This commit permits to have responsive font sizes on all the Odoo frontend. It is based on the Bootstrap mixin `font-size` which calculates the font size based on the viewport width.

Community PR: odoo/odoo#129469

task-1958098

### DIFF
--- a/theme_monglia/static/src/snippets/s_numbers/000.scss
+++ b/theme_monglia/static/src/snippets/s_numbers/000.scss
@@ -1,10 +1,4 @@
 .s_numbers {
-    .s_number {
-        font-size: 75px;
-        @include media-breakpoint-down(md) {
-            font-size: 60px;
-        }
-    }
     h6 {
         text-transform: uppercase;
     }


theme_common/static/src/scss/mixins.scss
```diff
@@ -68,7 +68,7 @@
         }
         .blockquote-text {
             overflow: hidden;
-            font-size: $font-size-base;
```
qsm-odoo, line 71: @Guillaume-gdi Let's not touch the theme_common/mixins.scss file. Those are old mixins only called in old snippets. If they still make sense, they can stay non responsive to further encourage not using this anymore.

theme_monglia/static/src/snippets/s_numbers/000.scss
```diff
@@ -1,8 +1,8 @@
 .s_numbers {
     .s_number {
-        font-size: 75px;
+        @include font-size(75px);
         @include media-breakpoint-down(md) {
-            font-size: 60px;
+            @include font-size(60px);
```
qsm-odoo, line 5: Using the font-size mixin and doing "if > md, use X else use Y" probably is useless, no? Doing `@include font-size(75px);`, I guess you should have something like 60px in md automatically, no?

qsm-odoo (changes requested):

theme_monglia/static/src/snippets/s_numbers/000.scss
Guillaume-gdi: I can even remove it because the .s_number has the display-4 class on it :+1:

theme_monglia/static/src/snippets/s_numbers/000.scss
```diff
@@ -1,11 +1,3 @@
-.s_numbers {
-    .s_number {
-        font-size: 75px;
-        @include media-breakpoint-down(md) {
-            font-size: 60px;
-        }
-    }
-    h6 {
+.s_numbers h6 {
         text-transform: uppercase;
```
Guillaume-gdi, line 2: @qsm-odoo During my check I saw that this uppercase rule doesn't work. Maybe we could juste remove it or we can create a task to fix it
image

qsm-odoo: As you wish, as long as it is done in a dedicated commit (maybe even PR) :+1:

Guillaume-gdi: Will be fixed in [this task](https://www.odoo.com/web#id=3469717&cids=1&menu_id=4720&action=333&active_id=974&model=project.task&view_type=form) :heavy_check_mark:

theme_monglia/static/src/snippets/s_numbers/000.scss
```diff
@@ -1,11 +1,3 @@
...
-        font-size: 75px;
-        @include media-breakpoint-down(md) {
-            font-size: 60px;
-        }
-    }
-    h6 {
-        text-transform: uppercase;
-    }
+.s_numbers h6 {
+    text-transform: uppercase;
```
qsm-odoo, line 2: Keep the `.s_numbers` selector alone. That way you'll preserve history but it's also a better convention for our snippet CSS files. (In general, it's good to regroup but I would keep the name of the snippet alone at the top).

qsm-odoo (changes requested):

qsm-odoo (approved):
@robodoo r+

PR #680, opened 2023-08-01
[FIX] theme_*: remove mb-2 class on orphan buttons

robinlej:
*: anelusia, artists, aviato, graphene, kiddo, loftspace, monglia,
notes, odoo_experts, paptic, test_custo, yes

The community PR [125824] introduces an automatic check on buttons to
add the `mb-2` class when several buttons follow each other, so that
they have space to breathe when split on multiple lines.
For consistency, this commit therefore removes most `mb-2` occurrences
on buttons within themes as long as they are alone on their line, but
keeps or adds the class when several buttons are next to each other.

[125824]: https://github.com/odoo/odoo/pull/125824

task-3369604


bvr-odoo:
@robodoo r+

bvr-odoo:
@robodoo r+

PR #682, opened 2023-08-01
[IMP] theme_*: prevent abuse of h1 and use font size classes

Guillaume-gdi:
This commit permit to reduce the risk to have a wrong h1 tag or multiple h1 tags on the same page. This commit replaces some h1 tags by a h2. In addition, this commit removed some font size written in the style attribute and replace them by a font size class.

Community PR: https://github.com/odoo/odoo/pull/129791

task-1958098


qsm-odoo (approved):
@robodoo r+

PR #685, opened 2023-08-04
[REF] test_themes: remove legacy patch

mcm-odoo:
This commit removes the use of legacy patch by using the modern patch.


jpp-odoo:
@robodoo r+

sdegueldre:
@robodoo r+

PR #686, opened 2023-08-09
[REF] test_themes: remove owl="1" from the templates

jpp-odoo:
As all the templates are now imported in the owl app, there is not need anymore to specify the owl="1" attribute in the templates.

Part of task~3443861


sdegueldre:
@robodoo r+

PR #688, opened 2023-08-17
[REM] theme_monglia: remove useless scss

Guillaume-gdi:
This commit removes the scss rule that makes the heading in the numbers block in uppercase. The rule was not working and is not very useful so this commit simply removes it.

task-3469717


Guillaume-gdi:
Wait for [this other PR](https://github.com/odoo/design-themes/pull/673) to be merged before merging this one

rdeodoo (approved):
@robodoo r+ thanks for the fix.

I rebased on https://github.com/odoo/design-themes/commit/6556a8cfff4be16b977b8d37c52d5f9ce819f58c

This is a dead file, you should have wrote that in the commit message, `The rule was not working and is not very useful so this commit simply removes it.` is not useful as you don't have any idea why.
Looking a second time at the diff, it's obvious since you don't remove it from the manifest, but I legit thought it was just a rule overridden by something else and spent a few minutes looking at the DOM style :)

I found the commit that removed the entry from the manifest, it's this [commit](https://github.com/odoo/design-themes-old/commit/14aad38f47a6d00f0d53e94debb8765dbbd81791#diff-5bc126dc3e67585d92c1ed55e9fe3093586830f492920be8a041b7c7bab70065L14) 
And it was initially (correctly) added with this [commit](https://github.com/odoo/design-themes-old/commit/69b99559275cfcfcd2b7f5d4e7667de44d9672b3)

I don't know if you know, but if you blame this line on the `design-themes` repo, you will end up on this ["commit"](https://github.com/odoo/design-themes/blame/14.0/theme_monglia/views/snippets/s_numbers.xml)
But you can actually go further in the history by replacing `design-themes` by `design-themes-old`, see the same file blame on the [old repo](https://github.com/odoo/design-themes-old/blame/14.0/theme_monglia/views/snippets/s_numbers.xml)

About master: it's safer indeed, in case someone fixed that locally by importing this css file or xml file.

The other solution was to make it work, but since `uppercase` is not something you can change in the builder.. Best not to.


I pushed to mention all that.

PR #641, opened 2023-04-06
[FIX] theme_*: adapt svg shapes paths

mano-odoo:
In commit https://github.com/odoo/odoo/pull/110673/commits/d82b2f6b8f9f5a56c470aa9c467b0fbb5ff7961a some of the shapes have been moved to new folder or have been renamed.

This commit adapt the paths of the previous shapes to their new location in various themes.

task-3266705

PR #689, opened 2023-08-17
[FIX] theme_anelusia: s_product_list broken img

Brieuc-brd:
Since PR #641, the url of item 4 was broken.

This commit fixes this issue.

### DIFF
--- a/theme_anelusia/views/snippets/s_product_list.xml
+++ b/theme_anelusia/views/snippets/s_product_list.xml
@@ -36,7 +36,7 @@
     </xpath>
     <!-- Item #4 - Img -->
     <xpath expr="(//img)[4]" position="attributes">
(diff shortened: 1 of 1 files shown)


Brieuc-brd:
Hello @qsm-odoo, this PR is ready for review, could you take a look at it ? Thanks !

qsm-odoo (approved):
Thanks for the fix :+1: Details for next time in the commit message:
- Try to mention merged commits instead of PR when referencing something that is merged
- The commit message should complete the sentence "When applied, this commit will ...", so here something like "restore a s_product_list image" would have been better than "s_product_list broken img".
- Messages like "This commit fixes this issue" are useless :wink: 

Like I said: details :) Not worth delaying the merge.

@robodoo r+

PR #692, opened 2023-08-21
[IMP] theme_*: make it possible to create new pages from templates

bso-odoo:
This commit removes the default shapes that are added on blocks by the themes in order to avoid having wrongly joined shapes when creating new pages from templates.

It also introduces customizations for the "New page from templates" feature

task-3381714

### DIFF
--- a/theme_anelusia/__manifest__.py
+++ b/theme_anelusia/__manifest__.py
@@ -44,7 +44,7 @@
         'website.s_media_list_default_image_1': '/theme_anelusia/static/src/img/snippets/s_media_list_1.jpg',
         'website.s_media_list_default_image_2': '/theme_anelusia/static/src/img/snippets/s_media_list_2.jpg',
     },
-    'snippet_lists': {
+    'configurator_snippets': {
         'homepage': ['s_text_cover', 's_images_wall', 's_color_blocks_2', 's_references',
                      's_media_list', 's_company_team', 's_call_to_action'],
     },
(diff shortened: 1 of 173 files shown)


qsm-odoo (changes requested):
@bso-odoo Let's talk about this tomorrow, live with @rdeodoo. Would be nice to have feedback of your meeting with the designers :)

For this `ignore_shape` stuff, I am not sure about the need. And if there is one, I don't think it should be about tweaking the definition of all theme overrides.

bso-odoo:
@rdeodoo Ouch... I just noticed the `new_page_templates.xml` for `theme_beauty` was not in the manifest... (it became obvious when merging the commits) So... I added it, it does not crash... but I think we will have to review the templates for that theme.

If we prefer, I can restore the previous display - by removing the customizations form that file. Sorry for not noticing earlier.

rdeodoo:
@robodoo rebase-ff

rdeodoo (approved):
@robodoo r+ rebase-ff

Impressive work, good job to all people related to those theme customization and Benoit for cleaning / adapting.
I wasn't expecting that deep customization / pixel perfect usage of the new page templates, and I am honestly not 100% fan of going that far but as it's 100% done according to the spec, it was the way to go.
Note that when I say not a fan, I mean it in a way that the technical debt is big here.
But since the added value is also awesome for our end users, I guess the `technical debt` / `added value` ratio is fine.

PR #693, opened 2023-08-29
[FIX] Update __manifest__.py to rename second image_2 to image_3

XtremXpert:
s_three_columns_default_image_2 was there twice, so I change the second one to s_three_columns_default_image_3

### DIFF
--- a/theme_avantgarde/__manifest__.py
+++ b/theme_avantgarde/__manifest__.py
@@ -20,7 +20,7 @@
         'website.s_picture_default_image': '/theme_avantgarde/static/src/img/pictures/bg_image_14.jpg',
         'website.s_three_columns_default_image_1': '/theme_avantgarde/static/src/img/pictures/bg_image_15',
         'website.s_three_columns_default_image_2': '/theme_avantgarde/static/src/img/pictures/bg_image_16.jpg',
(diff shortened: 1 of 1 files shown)


qsm-odoo (changes requested):
Hello :wave: 

Thank you for the fix :+1: Could you update the commit message to be more in line with our conventions? Something like that:
```
[FIX] theme_avantgarde: properly define 3rd 3-cols image for previews
```
- Max 72 characters per line
- [FIX] / [IMP] / [REF] to start
- Name of the impacted apps

In this case a description is not necessary. At best it could indicate the commit that introduced the bug. Something like that:
```
Since [1], the manifest is not well formed: it contains a duplicated key.

[1]: https://github.com/odoo/design-themes/commit/08befbeaf5ef90ef7ab2176847269d7eea2ec4c7
```

Thank you ! Note that once merged, the fix will automatically reach higher versions like 16.0, no need to mention it in the commit message either :+1:

qsm-odoo:
@XtremXpert Would you adapt the PR as suggested or do you prefer us handling the bug fix? Thank you! :+1:

qsm-odoo:
Merging at https://github.com/odoo/design-themes/pull/695

PR #695, opened 2023-09-21
[FIX] theme_avantgarde: properly define 3rd 3-cols image for previews

qsm-odoo:
Since [1], the manifest is not well formed: it contains a duplicated key

[1]: https://github.com/odoo/design-themes/commit/08befbeaf5ef90ef7ab2176847269d7eea2ec4c7

Courtesy of @XtremXpert


qsm-odoo (commented):
@robodoo r+

PR #703, opened 2023-09-28
[I18N] *: export 16.0 translations

mart-e:


rdeodoo:
Hello @mart-e 
Could you confirm that once this PR is merged, our translators will work on those new translations before OXP, ideally in main languages to reach close to 100% translation?
I believe JOPA meant that it would be the case but maybe I misread it: https://discord.com/channels/678381219515465750/1075751131797528616/1156930267781550110

For tracking purpose (https://discord.com/channels/678381219515465750/1075751131797528616/1156929303733350450), here is why we would really need a high % translated:
> We need those translated like 90+% in main languages for our ChatGPT integration to work
If we ask chatGPT to replace (generate new) content + translate it at the same time, and we only have partial translations, chatGPT just goes crazy.

mart-e:
@robodoo r+

> We need those translated like 90+% in main languages for our ChatGPT integration to work

Please don't give me arguments to remove all translations

mart-e:
>  Could you confirm that once this PR is merged, our translators will work on those new translations before OXP, ideally in main languages to reach close to 100% translation?

We can surely ask them to prioritize that part.

PR #711, opened 2023-10-12
[I18N] remove leftover modules

mart-e:


mart-e:
@robodoo r+
cc @oomsveta

PR #717, opened 2023-10-17
[IMP] themes_*: rewrite some messages and fix typos

rdeodoo:
Themes translation in Odoo 16 have been lost, they were not exported on Transifex anymore (bug).

We just added back the repository on it, and our internal translators are working on those translations.

They found some typos and bad wording which are fixed here.

Technically, note that we need the themes fully translated for the chatGPT integration coming in website configurator. Indeed, if a theme is not (almost fully) translated, it won't be using chatGPT for text replacement.
This is because if we ask chatGPT to do text replacement and translation at the same time, it just provides bad results.
See https://github.com/odoo/odoo/pull/137703


qsm-odoo (approved):
@robodoo r+

PR #719, opened 2023-10-17
[FIX] theme_*: adapt .pot file to recent commit changing wording

rdeodoo:
See commit [1], it changed wording but didn't adapt .pot files. As that commit was targetting master at first, .pot where not needed. When changing the target to odoo 16, we forgot to adapt the .pot.

[1]: https://github.com/odoo/design-themes/commit/7cdb6d2ab7054094b5651a43e3bce8e6c7e26e87


rdeodoo:
@robodoo r+

PR #726, opened 2023-10-17
[IMP] theme_*: restore base snippet shapes that do not connect blocks

bso-odoo:
In [1] all shapes were moved from the base snippet to the specific
configurator snippet. This was a bit overkill.

This commit reverts this partially so that only shapes that connect
with each other are specific to the configurator, but "intra-block"
shapes are restored onto the base snippet.
The "New page from template" snippets are adapted accordingly in order
to produce the same output as before this commit.

task-3555325

### DIFF
--- a/theme_anelusia/views/new_page_template.xml
+++ b/theme_anelusia/views/new_page_template.xml
@@ -3,50 +3,6 @@
 
 <!-- General customizations -->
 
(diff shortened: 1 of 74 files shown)


rdeodoo (approved):
@robodoo r+

In case someone is wondering, here is the result in image:

Before:
image

After:
image

We lost those "light" background images when drag & dropping blocks.

PR #731, opened 2023-10-23
[FIX] theme_*: remove duplicate and conflicting classes

bso-odoo:
When default snippets have been updated to rely on color presets ([1]),
some themes were not adapted accordingly.
This also happened whenever paddings of base blocks were modified.
Because of this there are elements in themes for which some classes are
applied several times, and some where classes are not removed before
adding a contradicting one.

This commit adjusts the classes so that themes neither:
- produce snippets with duplicated classes
- produce snippets with contradicting classes (e.g. two different
bottom paddings on the same element)

This commit also fixes a typo from [2] which was the source of
duplicate classes.

This PR also introduces a test to make sure the situation does not occur accidentally.

[1]: https://github.com/odoo/odoo/commit/5953c4df6b9c08519f64a65c1cf5d5558d59f7d4
[2]: https://github.com/odoo/design-themes/commit/299d504df31bf179775eec1d37e71605c3637460#diff-86c142838104de06e89f629c51cf791464538dc8899e6a3f00c24f69638db409R34

task-3562147

### DIFF
--- a/theme_bewise/views/new_page_template.xml
+++ b/theme_bewise/views/new_page_template.xml
@@ -51,9 +51,24 @@
     </xpath>
 </template>
 
+<template id="new_page_template_about_full_s_numbers" inherit_id="website.new_page_template_about_full_s_numbers">
+    <xpath expr="//section" position="attributes">
+        <!-- Added by both theme and new page template -->
+        <attribute name="class" add="o_cc3" remove="o_cc3" separator=" "/>
+    </xpath>
+</template>
+
+<template id="new_page_template_about_mini_s_cover" inherit_id="website.new_page_template_about_mini_s_cover">
+    <xpath expr="//section" position="attributes">
+        <attribute name="class" remove="pt40 pb40" separator=" "/>
+    </xpath>
+</template>
+
 <template id="new_page_template_about_personal_s_numbers" inherit_id="website.new_page_template_about_personal_s_numbers">
     <!-- Remove shape option -->
     <xpath expr="//section" position="attributes">
+        <!-- Added by both theme and new page template -->
+        <attribute name="class" add="o_cc3" remove="o_cc3" separator=" "/>
         <attribute name="data-oe-shape-data"/>
     </xpath>
     <!-- Remove shape -->
@@ -61,6 +76,12 @@
 </template>
 
 <!-- Snippet customization Landing Pages -->
+
+<template id="new_page_template_landing_0_s_cover" inherit_id="website.new_page_template_landing_0_s_cover">
+    <xpath expr="//section" position="attributes">
+        <attribute name="class" remove="pt200 pb200" separator=" "/>
+    </xpath>
+</template>
 
 <template id="new_page_template_landing_1_s_banner" inherit_id="website.new_page_template_landing_1_s_banner">
     <!-- Shape option -->
--- a/test_themes/tests/__init__.py
+++ b/test_themes/tests/__init__.py
@@ -2,4 +2,5 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from . import test_crawl
+from . import test_new_page_templates
 from . import test_theme_upgrade
--- a/test_themes/tests/test_new_page_templates.py
+++ b/test_themes/tests/test_new_page_templates.py
@@ -1,0 +1,208 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from lxml import html
+
+import logging
+import re
+
+from odoo.addons.website.tools import MockRequest
+from odoo.tests import tagged, TransactionCase
+from odoo.tools import escape_psql
+
+_logger = logging.getLogger(__name__)
+
+CONFLICTUAL_CLASSES = [
+    ['btn-outline-primary', 'btn-primary', 'btn-secondary'],
+    ['btn-block', 'btn-outline-primary'],
+    ['container', 'container-fluid', 'o_container_small'],
+    ['d-block', 'd-flex', 'd-inline-block', 'd-none'],
+    ['d-block', 'd-lg-block', 'd-md-block'],
+    ['d-flex', 'd-md-flex'],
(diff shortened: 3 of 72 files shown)


bso-odoo:
@rdeodoo This is the design-theme PR of https://github.com/odoo/odoo/pull/139572

rdeodoo (changes requested):
I am not going to review all the theme changes, honestly it's a stunning work and I'll trust you on every specific changes, I tried a few ones in theme_anelusia and it seems to be working as expected.

Thank you for the test, I left a few remarks about it.
I am more confident in having a strong test than reviewing line by line each XML diff here.

I am mitigated about introducing almost 200 new views but that's how the spec was designed, so..
For tracking purpose, we already had 1000 views related to `new_page_template_`, now it will be 1200. On a total of ~8500 views on a runbot design themes all modules installed (enterprise included).
But when you divide those 1200 views by ~30 themes, it's actually not that much to have that level of details in new page templates.

test_themes/tests/test_crawl.py
```diff
@@ -51,3 +147,96 @@ def test_02_homepage_tour_every_theme(self):
...
+                    ('key', 'like', f'{website.theme_id.name}.new_page'),
+                ])
+                for view in views:
+                    try:
+                        self.env['ir.qweb']._render(view.id)
+                    except Exception:
+                        errors.append("View %s cannot be rendered" % view.key)
+                        raise
+                view_count += len(views)
+        _logger.info("Tested %s views", view_count)
```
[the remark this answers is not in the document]
bso-odoo: I now use a set of ids in `test_render_templates`, but I kept a count in `test_render_applied_templates` - since it is always the same views that are rendered for each theme.

theme_bewise/views/new_page_template.xml
```diff
@@ -51,9 +51,24 @@
...
+    <xpath expr="//section" position="attributes">
+        <attribute name="class" remove="pt40 pb40" separator=" "/>
+    </xpath>
+</template>
+
 <template id="new_page_template_about_personal_s_numbers" inherit_id="website.new_page_template_about_personal_s_numbers">
     <!-- Remove shape option -->
     <xpath expr="//section" position="attributes">
+        <!-- Added by both theme and new page template -->
+        <attribute name="class" add="o_cc3" remove="o_cc3" separator=" "/>
```
rdeodoo, line 71: Remove is useless (or the entire line) -> same value in add and remove
There is 4 other occurences of this

bso-odoo: This is to avoid a duplicate `o_cc3`: `remove` removes all occurrences, `add` adds only the specified one.
(That's why I added a comment above those situations to mention that the class had been added from several templates)

test_themes/tests/test_crawl.py
```diff
@@ -51,3 +147,101 @@ def test_02_homepage_tour_every_theme(self):
...
+                            continue
+                        html_tree = html.fromstring(html_text)
+                        blocks_el = html_tree.xpath("//*[@id='o_scroll']")
+                        if blocks_el:
+                            # Only look at blocks in website.snippets
+                            html_tree = blocks_el[0]
+                        for el in html_tree.xpath('//*[@class]'):
+                            classes = el.attrib['class'].split(' ')
+                            classes_inventory.update(classes)
+                            if len(classes) != len(set(classes)):
```
rdeodoo, line 216:
```suggestion
                            if len(classes) != len(set(classes)):
```
```suggestion
                            if classes != set(classes):
```
Same?

rdeodoo (approved):
I moved the test myself and made the few little changes

rdeodoo:
@robodoo r+ rebase-ff

test_themes/tests/test_crawl.py
bso-odoo: No:
```py
>>> [1,2,3] == set([1,2,3])
False
```
Wrapping the `set` back inside a `list` works - but then my guess is that `len` is cheaper.

test_themes/tests/test_crawl.py
```diff
@@ -51,3 +147,101 @@ def test_02_homepage_tour_every_theme(self):
...
+                for view in views:
+                    try:
+                        self.env['ir.qweb']._render(view.id)
+                    except Exception:
+                        errors.append("View %s cannot be rendered" % view.key)
+        _logger.info("Tested %s views", len(view_ids))
+        self.assertGreater(len(view_ids), 1250, "Test should have encountered a lot of views")
+        self.assertFalse(errors, "No error should have been collected")
+
+    def test_render_applied_templates(self):
```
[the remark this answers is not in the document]
bso-odoo: Maybe there is a way to force the ORM to cache all the views in a single query before starting any rendering.
(I do not think there could be a way for us to determine which are the used ones.)

Note that on runbot it seems to be slightly faster (~72s).
```
2023-11-03 21:10:58,451 13 INFO 53300435-17-0-design_theme odoo.addons.test_themes.tests.test_new_page_templates: Starting TestNewPageTemplates.test_render_applied_templates ... 
2023-11-03 21:12:10,835 13 INFO 53300435-17-0-design_theme odoo.addons.test_themes.tests.test_new_page_templates: Tested 2929 views 
```
(from <https://runbot168.odoo.com/runbot/static/build/53300435-17-0/logs/design_theme.txt>)

test_themes/tests/test_crawl.py
rdeodoo: Indeed, glad I did not change that (test would have failed anyway)

test_themes/tests/test_crawl.py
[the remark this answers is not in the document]
rdeodoo: > Maybe there is a way to force the ORM to cache all the views in a single query before starting any rendering.

That exactly what I tried before merging the PR, but sadly it's not possible, the cache has no impact as it's an hard-coded query:
https://github.com/odoo/odoo/blob/342c37c3c593e85e2e9ebd5828d8dab3e7620842/odoo/addons/base/models/ir_ui_view.py#L599-L616

Regardless of the cache (eg caching the arch of every views in a single query), this is called.

PR #732, opened 2023-10-25
[FIX] theme_*: fix some configurator templates

bso-odoo:
Since [1] when background shapes were extracted to configurator-specific templates, some xpath were broken.
By separating the definition of the shape from the definition of the base snippet theme customization, changes between `container` and `o_container_small` are already applied before taking the shape definition into account.

This commit fixes the templates which have this issue.

task-3555325

### DIFF
--- a/theme_beauty/views/snippets/s_product_list.xml
+++ b/theme_beauty/views/snippets/s_product_list.xml
@@ -60,7 +60,7 @@
         <attribute name="data-oe-shape-data">{"shape":"web_editor/Wavy/15","flip":[]}</attribute>
     </xpath>
     <!-- Shape -->
-    <xpath expr="//div[hasclass('container-fluid')]" position="before">
+    <xpath expr="//div[hasclass('container')]" position="before">
         <div class="o_we_shape o_web_editor_Wavy_15"/>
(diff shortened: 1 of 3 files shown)


rdeodoo (approved):
@robodoo r+

LGTM, running this script output no errors (in `test_themes/test_crawl.py`):
```py
        def test_rde(self):
            errors = []
            view_count = 0
            from odoo.addons.website.tools import MockRequest
            websites_themes = self.env['website'].get_test_themes_websites()
                views = self.env['ir.ui.view'].search([
                    '|', '|',
                    ('key', 'like', f'{website.theme_id.name}.s_'),
                    ('key', 'like', f'{website.theme_id.name}.configurator'),
                    ('key', 'like', f'{website.theme_id.name}.new_page'),
                ])
                if not views:
                    print("Theme %r has no views??" % website.theme_id.name)
                with MockRequest(self.env, website=website):
                    for view in views:
                        try:
                            self.env['ir.qweb']._render(view.id)
                        except Exception as e:
                            print(e)
                            errors.append(view.key)
                view_count += len(views)
            print(view_count)
            print(errors)
```

PR #737, opened 2023-10-30
[FIX] theme_bewise: replace "diplomatic" by "academic"

bso-odoo:
This commit rephrases one of the quotes block's sentences to use the word "academic" instead of "diplomatic".


rdeodoo (approved):
@robodoo r+
Quick win, thanks !

Hopefully this will be translated in time for OXP

PR #745, opened 2023-11-20
[FIX] theme_*: Update homepage tours

hubvd:
Steps to reproduce
==================

Run the `test_02_homepage_tour_every_theme` tour.

Cause of the issue
==================

There is an homepage tour for every theme. Since the owl conversion of web_tour, none of the custom theme tour were run and for every website, the default homepage tour was used instead.

Some updates to those themes caused the tours to fail, but since they weren't run, no one noticed.

The issue has been fixed in community:
https://github.com/odoo/odoo/pull/142350

Solution
========

Since 1aa6700c986a96463165824400a9157c1db4e723,
some h1 tags were replaced by h2.

Since 68c68740bff7e96b9530de7d6fd4cd58145434a0,
The banner has been replaced by a text cover.

opw-3595512

### DIFF
--- a/theme_anelusia/static/src/js/tour.js
+++ b/theme_anelusia/static/src/js/tour.js
@@ -5,7 +5,7 @@ import wTourUtils from '@website/js/tours/tour_utils';
 const snippets = [
     {
         id: 's_text_cover',
-        name: 'Banner',
+        name: 'Text Cover',
     },
     {
         id: 's_images_wall',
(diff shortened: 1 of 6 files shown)


detrouxdev:
Hey @hubvd 👋🏻  I pushed a commit that should hopefully make the runbot green. If you want to squash the commits before merging that's fine by me 😄

hubvd:
It works, thanks :smiley:

PR #747, opened 2023-11-23
[FIX] theme_*: fix linter error for runbot build

rdeodoo:
There is a linter error in the modified files of this PR about `require` not defined.
It's actually not really a problem as those files are transcompiled from ES6 module to odoo module, injecting `require` before the file is actually used/read by the browser, at which point no issue actually arise.

Those linter error were introduced with PR [1] in Odoo 15 but were fix in Odoo 16 with commit [2] (which should have arguably be done in Odoo 15).

[1]: https://github.com/odoo/design-themes/pull/500
[2]: https://github.com/odoo/design-themes/commit/dc2f5112022b5f987ed418e04b5e15c93dc7a731

### DIFF
--- a/theme_artists/static/src/js/tour.js
+++ b/theme_artists/static/src/js/tour.js
@@ -2,9 +2,6 @@
 
 import wTourUtils from 'website.tour_utils';
 
-const core = require("web.core");
-const _t = core._t;
-
 const snippets = [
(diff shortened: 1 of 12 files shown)


rdeodoo:
@fw-bot up to saas-15.2

PR #751, opened 2023-11-29
[FIX] theme_*: register unused configurator snippets

bso-odoo:
When shapes were extracted to configurator snippets in [1], some configurator-specific snippets were created that do not appear on pages of the specific theme.
Because in [2] the call to `_generate_primary_snippet_templates` is done on a full list of themes instead of only the installed ones, the problem was not noticed: if any theme defines a block, it's website-side configurator-specific template is generated.
This is not the case during an upgrade: the templates are generated only for the installed themes. Because of this some "useless" configurator templates trigger an error when importing their XML definition because their parent template does not exist.

This commit fixes this by adding those templates in a pseudo page `_` in the `configurator_snippets` entry of each theme's manifest. In master, the templates will be removed instead.

[2]: https://github.com/odoo/odoo/commit/928eeca714a161f6bc03343e4dc8af9b050b9841#diff-f49a1e9eda23df9f1d48121ba376a5fabafe70ea18b29d4eab23d737e5d4eeb6R446

### DIFF
--- a/theme_anelusia/__manifest__.py
+++ b/theme_anelusia/__manifest__.py
@@ -48,6 +48,8 @@
     'configurator_snippets': {
         'homepage': ['s_text_cover', 's_images_wall', 's_color_blocks_2', 's_references',
                      's_media_list', 's_company_team', 's_call_to_action'],
+        # TODO In master, remove unused templates instead.
+        '_': ['s_comparisons'],
     },
     'license': 'LGPL-3',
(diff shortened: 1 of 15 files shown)


rdeodoo (approved):
@robodoo r+

I asked for a nightly single build per theme https://discord.com/channels/678381219515465750/704997651833159831/1180123592533483550

PR #755, opened 2024-01-08
[FIX] theme_*: call _generate_primary_templates before importing data

bso-odoo:
To make sure the required primary templates have been generated from the theme's manifest when inherited views are imported/updated, this commit triggers their generation from a first imported XML file.

This has been added to all themes for consistency, even the ones that do not require it currently.

Related to https://github.com/odoo/odoo/pull/148443

task-3670496

### DIFF
--- a/test_themes/tests/test_theme_standalone.py
+++ b/test_themes/tests/test_theme_standalone.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo.addons.website.tools import MockRequest
+from odoo.tests import standalone
+
+
+@standalone('website_standalone')
+def test_01_theme_install_generate_primary_templates(env):
+    """ This test ensures the theme `_generate_primary_snippet_templates()`
+    method is correctly called before xml views are generated.
+    """
+    # 1. Setup
+    theme_buzzy = env.ref('base.module_theme_clean')
+
+    if theme_buzzy.state == 'installed':
+        theme_buzzy.button_immediate_uninstall()
+    # Ensure those views are deleted to mimic the initial state of theme not installed.
+    # Because "theme_buzzy" was installed before through "test_themes" dependencies, removing
+    # those views is needed to replicate the bug: if the configurator views are not generated,
+    # the theme install will fail because some of the imported views inherit them.
+    env['ir.ui.view'].with_context(_force_unlink=True).search([('key', '=', 'website.configurator_s_banner')]).unlink()
+    env['ir.ui.view'].with_context(_force_unlink=True).search([('key', '=', 'website.configurator_s_cover')]).unlink()
+    theme_buzzy.button_immediate_install()
--- a/test_themes/tests/__init__.py
+++ b/test_themes/tests/__init__.py
@@ -3,4 +3,5 @@
 
 from . import test_crawl
 from . import test_new_page_templates
+from . import test_theme_standalone
 from . import test_theme_upgrade
--- a/theme_anelusia/__manifest__.py
+++ b/theme_anelusia/__manifest__.py
@@ -7,6 +7,7 @@
     'version': '2.1.0',
     'depends': ['theme_common'],
     'data': [
+        'data/generate_primary_template.xml',
         'data/ir_asset.xml',
         'views/images_content.xml',
         'views/images_library.xml',
--- a/theme_anelusia/data/generate_primary_template.xml
+++ b/theme_anelusia/data/generate_primary_template.xml
@@ -1,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<!-- Generate primary snippet templates that are not predefined -->
+<function model="ir.module.module" name="_generate_primary_snippet_templates">
+    <value eval="[ref('base.module_theme_anelusia')]"/>
+</function>
+
+</odoo>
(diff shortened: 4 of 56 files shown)


test_themes/tests/test_theme_standalone.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
+
+from odoo.addons.website.tools import MockRequest
+from odoo.tests import standalone
+
+
+@standalone('website_standalone', 'test_rde')
```
rdeodoo, line 8:
```suggestion
@standalone('website_standalone')
```

rdeodoo (changes requested):
cc @qsm-odoo for point 2.

The fix is LGTM, but I added 2 commits:
1. One to introduce a test to cover the fix you made, it's really a very critical fix, not having this kind of flow covered is really not an option, especially once we have detected a failure.
It comes in a form of a `standalone` test. I don't know if you are familiar with it or not, there is probably a very few people knowing about it so.. Just let me know, I added some docs in our internal discord back in the days, there is a few info there but we can discuss it if needed.
Could you have a look at the test and confirm it's looking good to you? After that I'll be able to merge it :+1: 
2. One to fix a mistake made in the themes, which should probably be backported :grimacing: : some themes were not depending of theme_common, which shouldn't be that big of a deal but still wrong. On top of what I wrote in the commit msg, it's also making `test_themes` acts weirdly: it's not going through the `_post_copy` of those themes when installing those through its `__init__.py` file.
It's probably a stable friendly change to make.

rdeodoo:
> I also think it's not stable friendly to add a dependency to a module as this requires a migration script, no?

Indeed, this is not stable friendly, I still needed to test before validating this one:
- Install website
- Select theme buzzy
- Apply new dependency diff
- Restart server
-> Website error

> If I am not wrong, I already see a very big "ir_asset.xml" file in theme_common which defines ir.asset records of compatibility CSS to enable... but that file is not even mentioned in the manifest 🤦‍♂️

Yep.. Seems like an issue when opensourcing the themes, probably failed some commit operation.

> And theme_common is only about compatibility stuff, there is nothing else in there anymore.

What about this (which I mentioned in the commit msg): https://github.com/odoo/design-themes-old/commit/50b3b1bd03658f0a49524268c41281704d231a1b
Isn't it something we still need/want?

But on top of that, isn't there other issues when not having theme_common as dependency? I guess no otherwise we would've figured, but when installing `test_themes`, without this "theme_common" as a dependency, those themes are not behaving as others.
I am not sure exactly:
- I initially wrote a test here with `button_choose_theme`
- It was not working as expected with those themes because of the theme_common not being a dependency -> You can see it in the logs when it loops through all themes in test_themes `__init.py__` to create a website per theme, those themes-website doesn't go through the full process, I believe their views xml file were ignored or something like that, because of `self` being an empty record in `_theme_upgrade_upstream`, while it was the theme record for all other themes (`themes.filtered(lambda m: m.state == 'installed').button_upgrade()` -> it was not 'installed')
- I actually fixed it, made the test work, then realized it was even simpler to cover the issue being fixed in this PR: just call `button_immediate_install`, no need to apply it on a website
- With this new way of writing the test, the "theme_common" in manifest missing is not an issue anymore

That's for the whole story of this fix. It's still a bit concerning because `test_themes` is not doing too much magic, it basically calls `_theme_load` in its `__init.py__`, not too far from what the codebase really does when switching theme in the UI.

IMHO it was not ideal to not add theme_common as dependency, just for the sake of consistency and possible bugs, at least until we fully get rid of it.

We can probably just ignore it, it's not worth time to recheck everything (took me a lot of hours to write this test, partly because of the theme_common issue). Just keep this in mind should we have some issues with those themes one day.

rdeodoo (approved):
@robodoo r+

qsm-odoo:
>  it was not ideal to not add theme_common as dependency, just for the sake of consistency and possible bugs, at least until we fully get rid of it.

Maybe but I don't agree. `theme_common` was a bunch of utils before, that's why themes were adding it as dependency. It stopped being a bunch of utils, only compatibility stuff, new themes had thus no business in requiring it. As for the "possible bugs" it would cause not depending on it, I guess it's like every code we write :shrug: To my knowledge, nothing was mentioning theme_common as part of a requirement for things to work. Before we also had `theme_XXX_sale` and others, the system was thus supposed to work with a whole hierarchy of themes to install, theme_common was not an exception to that system IMO

> What about this (which I mentioned in the commit msg): https://github.com/odoo/design-themes-old/commit/50b3b1bd03658f0a49524268c41281704d231a1b

Maybe, but: I would be surprised if those compatibility code solves more problems than they create. Those views that are disabled are probably removed by upgrade script at this point... or users who still have them left Odoo :shrug: We should really put more effort into removing theme_common at this point. I don't know how. Now that we have full control over upgrade scripts, I'd say it would be time to just write some for good conscience (e.g. removing the views the commit you mentioned handles... provided they still exist in the user database) and remove the rest knowing it could break one or two snippets in one or two databases over thousands :shrug: Maybe we actually can write a script about that too: "if that compatibility view exist in the DB, just remove it and mark all snippets they targeted as outdated". TBH at this point, even just "drop theme_common without upgrade script" is probably a valid way to go... if we did not care as we do :grimacing: 

> Seems like an issue when opensourcing the themes, probably failed some commit operation.

Was not that already version 16.0 then? And in that case, should not all upgrade fail if users still had those views in 15.0 (as no upgrade script is removing them). To investigate... maybe we'll actually find a reason to be able to remove theme_common without much upgrade script.

rdeodoo:
@robodoo rebase-ff

rdeodoo:
https://www.odoo.com/web#menu_id=4720&cids=1&action=333&active_id=974&model=project.task&view_type=form&id=3696867

PR #756, opened 2024-01-10
[FIX] theme_test_custo: consider the `top_menu` navbar class in tests

sobo-odoo:
In PR [1], the duplicated navbars ids `o_main_nav` and `top_menu` are
replaced by classes, to avoid having the same ids multiple times in the
DOM. Note that for stability, the ids are kept for the desktop view
navbar.

This commit adapts the tests so they also consider the classes and not
only the ids.

[1]: https://github.com/odoo/odoo/pull/146492

task-3609531

See PR [1] for the associated community PR.


qsm-odoo (approved):
@robodoo r+

PR #762, opened 2024-01-29
[IMP] test_themes: test that theme_default templates are generated

bso-odoo:
This commit adds a test that verifies if the theme_default's configurator templates are generated whenever the website module is either installed or updated.

Related to https://github.com/odoo/odoo/pull/151501.

task-3670496


rdeodoo (approved):
@robodoo r+

PR #766, opened 2024-02-09
[IMP] theme_*: POC website design in db

Guillaume-gdi:
This PR is a walking skeleton of the website design in db.

task-2463604


theme_avantgarde/static/src/scss/primary_variables.scss
```diff
@@ -177,7 +177,6 @@ $o-theme-font-configs: (
 $o-website-values-palettes: (
     (
         'color-palettes-name': 'avantgarde-3',
-        'layout': 'full',
```
qsm-odoo, line 180: Don't hesitate to make extra PR that I can merge directly for things like these: defining layout = full in a theme is useless since this is a default (this is not wrong but I'd say it is best not to write it if useless).

If I can merge small stuff like these that are independent of your work directly, it will make the review easier later :+1:

theme_graphene/data/theme_website_design.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<record id="theme_graphene.design" model="theme.website.design">
+    <field name="layout">'boxed'</field>
+    <field name="theme_id" ref="base.module_theme_graphene"/>
```
qsm-odoo, line 6: You are actually filling the "theme.utils" many2one with a "ir.module.module" record, funny if it works ^^ Anyways, like I said in the other PR: you don't need that. You are in theme_graphene, you should never have to say "I am a theme_graphene record".

qsm-odoo (changes requested):

PR #779, opened 2024-02-29
[FIX] test_themes: fix test_01_theme_upgrade_post_copy

detrouxdev:
Commit [1] introduced a new test that uninstalls theme_buzzy. This leads to the whole module being uninstalled because this module has all themes as dependencies.

To fix this, we just run the test as the last test, ensuring the module is still installed for other tests.

Ideally, test would be independent from one another, but for now, the order matters.

[1]: https://github.com/odoo/design-themes/commit/4cdd1db20c021ebd369b46c739591def9732eec5

### DIFF
--- a/test_themes/tests/__init__.py
+++ b/test_themes/tests/__init__.py
@@ -1,7 +1,11 @@
 # -*- coding: utf-8 -*-
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
+# The order of these tests is important, as the last one will uninstall this
+# module. Ideally, the tests would be independent from one another but for now,
+# ensure this order.
 
 from . import test_crawl
 from . import test_new_page_templates
-from . import test_theme_standalone
 from . import test_theme_upgrade
+# This test should be last.
+from . import test_theme_standalone


qsm-odoo:
@robodoo delegate=sobo-odoo 
@sobo-odoo

test_themes/tests/__init__.py
```diff
@@ -1,7 +1,11 @@
 # -*- coding: utf-8 -*-
```
sobo-odoo, line 1: Just a question (I am not good with all the modules and python stuff :sweat_smile:): at the end of the problematic test, the buzzy module is reinstalled right ? So shouldn't the `test-themes` one have been reinstalled with it too ? (I am not sure on how the dependencies work)
If not, then why not reinstalling this module manually too (with `button_immediate_install()`) instead of changing the order of the tests ? :slightly_smiling_face:

sobo-odoo, line 1: Commit message:
- `Ideally, test would` -> tests

sobo-odoo (changes requested):
I think the fix is alright :+1: The explanation makes sense and since it is just changing the order, there is no risk in approving the PR (I am not sure on how to test the whole flow like on runbot, can you tell me ?)

I just have a question and a small comment on the commit message (see review comment) :slightly_smiling_face:

test_themes/tests/__init__.py
detrouxdev: Good question! While the theme is reinstalled, test_themes is not. Reinstalling test_themes causes a bunch of errors (most likely cause test_themes creates a lot of records). Since test_themes is not a module installed by users, this fix is faster and probably safer too.

sobo-odoo (approved):
Alright, let's r+ ! :wink: (you can resolve the conversation, I apparently don't have the button to do it :shrug: )

@robodoo r+

rdeodoo:
As discussed, at some point the standalone tests should be split so they are really standalone, otherwise it will be impossible to keep introducing some.
See https://discord.com/channels/678381219515465750/704997651833159831/1212739417295429692
Worth ensuring they will do it, I'll keep an entry in my TODO list.

PR #783, opened 2024-03-12
[REF] *: use iframe hoot pseudo selector in tour triggers

pipu-odoo:
In this commit, we change rest of triggers in tours where an iframe is selected (suite of #156653).
To do this, we use the :iframe HOOT pseudo-selector. With this changes, we can now remove the getNodesFromSelector function in odoo/addons/web_tour/static/src/tour_service/tour_utils.js Note that in this previous function, a ([is-ready]="true") check was made for iframe elements but was only necessary for few of them (7). So, a stepUtils.waitIframeIsReady() has been added to check this state.


aab-odoo:
robodoo r+

sdegueldre:
@robodoo delegate=Arcasias

PR #784, opened 2024-03-15
[FIX] theme_*: make s_picture image responsive by default

qsm-odoo:
Some themes were making the s_picture image responsive on their own. It is now done generically, so the theme extensions must be removed to be able to merge.

task-3266862


qsm-odoo (commented):
@robodoo r+

PR #793, opened 2024-03-26
[I18N] update .tx/config project

ticodoo:
config file was still pointing to 17 project, so in order for it to appear in the saas-17.1, we need to update it. Note that .pot files had any difference since previous export so they are left unchanged (i.e. the `Project-Id-Version` doesn't matter)


oomsveta:
robodoo r+

PR #798, opened 2024-04-09
[REM] theme_common, theme_*: remove theme_common

bso-odoo:
Theme common used to contain tools that were reused by most of the themes. It has now become an empty module which pollutes the module dependency tree.

This commit removes `theme_common` and adapts themes that used to depend on it.

task-3696867


bso-odoo:
Related upgrade PR: https://github.com/odoo/upgrade/pull/5906

PR #805, opened 2024-06-11
[REF] theme_*: convert default images to WebP format

SergeBayet:
This commit switches the default image format from JPEG to WebP for all design-themes images. WebP offers superior compression compared to JPEG, resulting in smaller file sizes and faster website loading times. Additionally, this change helps avoid warnings from performance tools regarding image optimization.

See also: https://github.com/odoo/odoo/pull/168862

task-3167548


qsm-odoo (changes requested):
Will need a rebase too of course :pray:

PR #815, opened 2024-06-26
Fix typo in fr translation

techjpiron:


qsm-odoo:
Hello :wave: 

Thank you for your fix. However, translating strings is done via Transifex (you can learn more about that here: https://github.com/odoo/odoo/wiki/Translations). I suggested your nice change in there, it just got approved. Translated strings from Transifex are synchronized every Sunday on GitHub, so you can expect this to be fixed next week.

Best regards

PR #816, opened 2024-07-01
[REF] theme_* : remove extra_trigger

pipu-odoo:
In order to simplify the tours API, it was decided to remove the extra_trigger key from the steps.
To check that an element is in the DOM, simply create a step with a trigger.

task~3974087

https://github.com/odoo/odoo/pull/171309


qsm-odoo (changes requested):
Commit message nitpicking: there is no "web_tour" in the themes repo, you can just put `*` or `theme_*`

PR #831, opened 2024-07-16
[I18N] export source terms

oomsveta:


theme_common/i18n/theme_common.pot
```diff
@@ -4,10 +4,10 @@
 #
 msgid ""
 msgstr ""
-"Project-Id-Version: Odoo Server 17.0+e\n"
+"Project-Id-Version: Odoo Server saas~17.4\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-09-28 12:25+0000\n"
-"PO-Revision-Date: 2023-09-28 12:25+0000\n"
+"POT-Creation-Date: 2024-07-15 12:24+0000\n"
+"PO-Revision-Date: 2024-07-15 12:24+0000\n"
```
ticodoo, line 10: `theme_common` isn't in the .tx/config (looks like it never has been). Not sure if we need it though, it's only the term: `Theme Utils`, i.e. the class description :shrug:

ticodoo (approved):

theme_common/i18n/theme_common.pot
oomsveta: There were some restrictions on the number of entries to prevent files with less than 3 entries from being listed in .tx/config. I don't know why it was like that, I removed it when exporting Enterprise and Community, but apparently not for Design Themes. I think this is the explanation 👍

oomsveta:
robodoo r+

PR #837, opened 2024-07-19
[FIX] theme_kiddo: remove vimeo video that no longer exists

detrouxdev:
The video used in theme_kiddo for the banner snippet no longer works. Either because it was requested too many times too quickly, or because the video was removed from Vimeo.

This commit removes the video. It does not replace it because pulling informations from external ressources which aren't guaranteed or handled by Odoo tend to randomly crash the Runbot.

This is actually the reason for this fix in the first place. Staging is crashing for 15.0.

runbot-71672


detrouxdev:
@robodoo fw=no

Xavier-Do:
@robodoo r+

Xavier-Do:
@robodoo priority

PR #839, opened 2024-07-22
[IMP] theme_bookstore, *: adapt to new `s_masonry_block` design

stefanorigano:
*: theme_notes, theme_real_estate, theme_yes, theme_zap

Community:
- https://github.com/odoo/odoo/pull/173870

task-3665303
part of task-3619705


theme_yes/views/snippets/s_masonry_block.xml
```diff
@@ -2,46 +2,26 @@
...
 
     <!-- Little block #02 - Title -->
     <xpath expr="(//h3)[2]" position="replace" mode="inner">
         Venues
     </xpath>
     <!-- Little block #02 - Text -->
-    <xpath expr="//*[hasclass('col-lg-3')]//p" position="replace">
+    <xpath expr="//*[hasclass('col-lg-4')][2]//p" position="replace">
         <p>Check out our list of favorite venues.</p>
         <p><a href="#" class="btn btn-secondary">Find a venue</a></p>
```
sobo-odoo, line 23: It seems that the texts of "Attire" and "Venues" have been inverted :+1: 
image

Surely because the second "h3" is not in the second "col-lg-4" (which is why the order of the elements is important (see my comment about the mobile order in the community PR) :wink: 

By the way, I've always found this image to be a bit of bad quality (like it seems pixelated) => maybe you could replace it (just saying) ? :slightly_smiling_face:

theme_bookstore/views/snippets/s_masonry_block.xml
```diff
@@ -4,70 +4,69 @@
 <template id="s_masonry_block_default_template" inherit_id="website.s_masonry_block_default_template">
```
sobo-odoo, line 4: You forgot the customization in the `theme_vehicle/.../customization.xml` file :wink: 
You should set the container as `container-fluid` and it seems to not target the right elements when changing the texts :+1: 
image

sobo-odoo (changes requested):
Here is a small review :slightly_smiling_face:

theme_vehicle/views/customizations.xml
```diff
@@ -82,19 +82,19 @@
...
     </xpath>
+    <xpath expr="//section/div" position="attributes">
+        <attribute name="class" add="container-fluid" remove="container" separator=" "></attribute>
+    </xpath>
 </template>
 
 <template id="s_masonry_block_image_texts_image_template" inherit_id="website.s_masonry_block_image_texts_image_template" name="Vehicle s_masonry_block">
-    <xpath expr="//div[hasclass('col-lg-3')]" position="replace" mode="inner">
-        <img src="/web/image/website.s_parallax_default_image" class="img img-fluid mx-auto" alt=""/>
-    </xpath>
```
sobo-odoo, line 90: With the parallax image, it was looking like that before 
image

But now, since there is no second masonry block image, it defaults to the default one and looks like this
image

Maybe it is worth keeping the parallax image ? :slightly_smiling_face:

sobo-odoo (changes requested):
Small detail to fix :wink:

theme_vehicle/views/customizations.xml
stefanorigano: Ooops, thank you, fixed 👍

stefanorigano:
@sobo-odoo,  "vehicle" detail fixed and  "yes" pixelated  image replaced 👍 
Branch rebased, thanks!

sobo-odoo (approved):
@robodoo r+

PR #866, opened 2024-08-13
[IMP] theme_*: adapt themes with redesigned `s_banner` snippet

anso-odoo:
*: (buzzy, cobalt, enark, kiddo, real_estate, zap)

This commit adapts the design of new `s_banner` snippet for multiple themes.

requires: https://github.com/odoo/odoo/pull/176595

task-4109584

### DIFF
--- a/theme_buzzy/views/snippets/s_banner.xml
+++ b/theme_buzzy/views/snippets/s_banner.xml
@@ -11,14 +11,15 @@
    <!-- Remove grid images -->
    <xpath expr="//div[hasclass('o_grid_item_image')]" position="replace"/>
    <xpath expr="//div[hasclass('o_grid_item_image')]" position="replace"/>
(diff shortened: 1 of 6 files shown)


sobo-odoo:
@anso-odoo I [force pushed](https://github.com/odoo/design-themes/compare/b9a76d881654983098ad6063f6123535718d7c07..dae4b63c1945df94b4c06fb90edb5b525407bbef) to:
- also remove the `data-row-count` attribute when removing the grid mode
- fix the xpath in enark (wrong `g-height/col` classes)
- remove the `h-100` class from real estate (already there before) => this should never be used, firstly because as usual there is no way to add/remove it in the editor, secondly because it had no real effect (the column already takes the whole space) and third, because when toggling the grid mode, it was completely broken, since it wants to take 100% of the space :grimacing:

sobo-odoo (approved):
@robodoo r+

PR #887, opened 2024-08-28
[IMP] theme_test_custo: adapt portal.user_dropdown classes

Brieuc-brd:
Since `btn` has been removed by default from `portal.user_dropdown` classes, this commit adapts the template call to maintain the correct behavior.

task-4019055

Requires:
- https://github.com/odoo/odoo/pull/178150


qsm-odoo (changes requested):
@Brieuc-brd I'll r+ the community one but this one should just be closed right? :thinking: (you can do it if you agree)

Brieuc-brd:
Indeed, this one is no longer necessary, thanks @qsm-odoo 👍

PR #889, opened 2024-08-29
[IMP] test_themes: add test to detect accidental cross-theme references

bso-odoo:
This commit adds a test to ensure that assets and images referenced from
a theme do actually belong to that theme.

task-4146886

### DIFF
--- a/theme_anelusia/views/images_library.xml
+++ b/theme_anelusia/views/images_library.xml
@@ -20,7 +20,7 @@
 <record id="s_text_cover_default_image" model="theme.ir.attachment">
     <field name="key">website.s_text_cover_default_image</field>
     <field name="name">website.s_text_cover_default_image</field>
-    <field name="url">website.s_banner_default_image</field>
+    <field name="url">/theme_anelusia/static/src/img/snippets/s_banner.jpg</field>
 </record>
 <record id="s_popup_default_image" model="theme.ir.attachment">
     <field name="key">website.s_popup_default_image</field>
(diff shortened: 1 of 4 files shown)


theme_anelusia/views/images_library.xml
```diff
@@ -20,7 +20,7 @@
 <record id="s_text_cover_default_image" model="theme.ir.attachment">
     <field name="key">website.s_text_cover_default_image</field>
     <field name="name">website.s_text_cover_default_image</field>
-    <field name="url">website.s_banner_default_image</field>
+    <field name="url">/theme_anelusia/static/src/img/snippets/s_banner.jpg</field>
```
detrouxdev, line 23: As I said in DMs, I wonder if this is a wrong URL. We'll have to look up whether this is ok to have or not, because I assume your test flagged it as wrong? In any case, it's not bad if we enforce some uniformity to those records anyway.

detrouxdev (approved):
It seems like the lines that were wrong were introduced in previous commits that are on stable branches:
- [theme_anelusia (17.0)](https://github.com/odoo/design-themes/commit/68c68740bff7e96b9530de7d6fd4cd58145434a0)
- [theme_artist (15.0)](https://github.com/odoo/design-themes/commit/30c3b2cdae4c6eae1f672259a5f2e5c7cba58132)

However, fixing in stable would also require another solution, so if needed, we can backport this later with a solution that works for stable.

This also doesn't seem to cause any issue when dropping the snippets, except if someone tries to isolate the themes inside the design-themes repo. Let's assume no one is doing that for now.

I'm merging this now as we have a lot of changes to snippets that could potentially introduce new errors like this and the test could catch them.

@robodoo rebase-ff r+

PR #892, opened 2024-08-30
[IMP] theme_*: adapt `s_company_team_basic`

mano-odoo:
*: artists, bewise, buzzy, notes, yes, zap

This commit adapts the design themes to the new `s_company_team_basic` snippet.

task-4144648
Part of: task-4077427

Community: https://github.com/odoo/odoo/pull/178323

### DIFF
--- a/theme_anelusia/views/snippets/s_company_team_basic.xml
+++ b/theme_anelusia/views/snippets/s_company_team_basic.xml
@@ -1,0 +1,79 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<template id="s_company_team_basic" inherit_id="website.s_company_team_basic">
+    <!-- Team #01 - Title -->
+    <xpath expr="//div[hasclass('o_grid_mode')]/div[2]/h4" position="replace" mode="inner">
+        Tony Fred
+    </xpath>
+    <!-- Team #01 - Work Description -->
+    <xpath expr="//div[hasclass('o_grid_mode')]/div[2]/p[last()]" position="replace" mode="inner">
+        Editor-in-chief
+    </xpath>
+    <!-- Team #01 - Img -->
+    <xpath expr="(//img)[1]" position="attributes">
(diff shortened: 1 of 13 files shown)


theme_anelusia/views/snippets/s_company_team_basic.xml
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<template id="s_company_team_basic" inherit_id="website.s_company_team_basic">
+    <!-- Team #01 -->
```
anso-odoo, line 5: The structure is a bit weird, shouldn't this change be under `<!-- Team #01 - Img -->` (same for other team members)

anso-odoo, line 5: Should this change be under `<!-- Team #01 - Img -->` ? Same for other team members

anso-odoo (approved):
Hey @mano-odoo can you add all the themes you are targeting in the PR description ?

Also a small comment but after that, LGTM

theme_anelusia/views/snippets/s_company_team_basic.xml
mano-odoo: Woops good catch I don't need to repeat the expression, it's could be part of the same xpath 👍

theme_anelusia/views/snippets/s_company_team_basic.xml
```diff
@@ -0,0 +1,79 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<template id="s_company_team_basic" inherit_id="website.s_company_team_basic">
```
SergeBayet, line 4: Consistency. Some inherited (see Be wise theme) templates have a name attribute and some not.

SergeBayet (changes requested):
LGTM, some nitpicking however. See below.

SergeBayet (approved):

PR #895, opened 2024-09-04
[FIX] theme_*: fix font style of snippet headers

bvr-odoo:
Some snippets contain incorrect font style that are not achievable using the editor (for example: `<h2 class="display-3">` in the "Key benefits" snippet).

These snippets were adapted in the community commit linked to this one to align with what the editor would generate if we recreated them from scratch.

In this commit, we adapt the themes following the change in community. We also fix a few issues (e.g., wording for the ´s_numbers´ snippet) noticed during the changes.

task-4148057


bso-odoo:
@robodoo delegate=robinlej

robinlej:
@robodoo r+

PR #915, opened 2024-09-17
[FIX] theme_bewise, *: adapt `s_image_title` without `display: flex`

sobo-odoo:
*: vehicle

In the community PR, the `s_image_title` snippet has been modified in order to not be in `display: flex` anymore, as it could cause issues with images. The same visual was obtained by setting a `grid-area` in a way that the text block is at the bottom right.

This commit adapts this snippet customizations to these changes. More precisely, it adapts the `grid-area` and padding to the text content, in order to avoid having an overflow while still looking like before.

Related to task-4105319


qsm-odoo (approved):
@robodoo rebase-ff r+

qsm-odoo:
@robodoo r+

PR #920, opened 2024-09-19
[IMP] theme_bookstore: theme improvement

anso-odoo:
Revamp of `theme_bookstore`

task-4178030


qsm-odoo:
Being moved to https://github.com/odoo/design-themes/pull/969 (hopefully, with correct Signed-off-by)
Don't push here or there anymore, if you need to apply fixes, create a new PR to merge once the main one is merged.

theme_bookstore/__manifest__.py
```diff
@@ -58,25 +58,27 @@
...
-        'website.s_product_list_default_image_6': '/theme_bookstore/static/src/img/snippets/s_product_6.jpg',
+        'website.s_banner_default_image_3': '/theme_bookstore/static/src/img/snippets/s_banner_3.jpg',
+        'website.s_image_text_default_image': '/theme_bookstore/static/src/img/snippets/s_image_text.jpg',
+        'website.library_image_02': '/theme_bookstore/static/src/img/snippets/library_image_02.jpg',
+        'website.s_cover_default_image': '/theme_bookstore/static/src/img/snippets/s_cover.jpg',
+        'website.s_media_list_default_image_1': '/theme_bookstore/static/src/img/snippets/s_media_list_1.jpg',
+        'website.s_image_text_default_image': '/theme_bookstore/static/src/img/snippets/s_image_text.jpg',
+        'website.s_media_list_default_image_2': '/theme_bookstore/static/src/img/snippets/s_media_list_2.jpg',
+        'website.s_text_image_default_image': '/theme_bookstore/static/src/img/snippets/s_text_image.jpg',
     },
```
sobo-odoo, line 79: Maybe some things to check here, after investigating my comment on the theme svg :slightly_smiling_face:

sobo-odoo (commented):
Given that it was merged without a deep review, here is my review :slightly_smiling_face: 
Lots of nitpicks about shapes that make sense only in the configurator context, and which would therefore be better to be added on `configurator_s_...` templates :+1: But not really necessary, if you judge they look good alone, it can stay like that :slightly_smiling_face: 

I also have comments about the svg preview images, which I think are wrongly set, but I don't really know about it so tell me if it was intentional :slightly_smiling_face:

sobo-odoo:
@anso-odoo (forgot to ping, just in case :slightly_smiling_face: )

PR #933, opened 2024-09-24
[IMP] test_themes: verify separators used for attributes

bso-odoo:
This commit adds a test that verifies that a specific `separator` is
used when an `attribute` tag is used with either `add` or `remove`.

task-4206845

### DIFF
--- a/test_themes/tests/test_new_page_templates.py
+++ b/test_themes/tests/test_new_page_templates.py
@@ -190,6 +190,12 @@ class TestNewPageTemplates(TransactionCase):
                                 conflict.difference_update(white_list)
                                 if len(conflict) > 1:
                                     errors.append("Using %r, view %r contains conflicting classes: %r in %r (according to pattern %r)" % (theme_name, view.key, conflict, classes, conflicting_classes_re.pattern))
+                        for el in html_tree.xpath('//*[@style]'):
+                            styles = el.attrib['style'].split(';')
+                            non_empty_styles = filter(lambda style: style, styles)
+                            property_names = list(map(lambda style: style.split(':')[0].strip(), non_empty_styles))
+                            if len(property_names) != len(set(property_names)):
+                                errors.append("Using %r, view %r contains duplicate style properties: %r" % (theme_name, view.key, el.attrib['style']))
                     except Exception:
                         _logger.error("Using %r, view %r cannot be rendered", theme_name, view.key)
                         errors.append("Using %r, view %r cannot be rendered" % (theme_name, view.key))
--- a/test_themes/tests/test_new_page_templates.py
+++ b/test_themes/tests/test_new_page_templates.py
@@ -1,6 +1,6 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
-from lxml import html
+from lxml import etree, html
 
 import logging
 import re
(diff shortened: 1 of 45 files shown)


agau-odoo (changes requested):
Thanks for the PR! :sunglasses: 

Here is a review with a few comments for your first 2 commits, mostly nitpicking/codestyle.

Your test works well for me, locally (red/green). :white_check_mark: 

The fix for your test case cannot come after the introduction of the test itself, otherwise the latter fails. I would consider swapping or squashing them.

Are we allowed to use this formatting trick for module names? `theme_*` ... `*: a, b ,c`
I really hope we are, or some variant, but imho we need to check.

test_themes/tests/test_new_page_templates.py
```diff
@@ -210,3 +210,44 @@ def check(theme_name, website):
...
+            ])
+            for view in views:
+                try:
+                    xml_tree = etree.fromstring(view.arch_db)
+                    els = xml_tree.xpath('//attribute[@add] | //attribute[@remove]')
+                    for el in els:
+                        name = el.attrib['name']
+                        if name in ATTRIBUTE_SEPARATORS:
+                            separator = el.attrib.get('separator', ',')
+                            expected_separators = ATTRIBUTE_SEPARATORS[name]
```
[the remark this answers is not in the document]
bso-odoo: In the context of tests, I'd rather stick to the naming "expected", given that the classic way of using test frameworks is: `assertSomething(expectedValue, actualValue, furtherMessage)`.
I know it is a bit abused here with a global `assertFalse` - but the word expected was used on purpose (even though it is a set of possible values here).
But if you insist on this one, I'll change it ^_^

bso-odoo:
@robodoo rebase-ff

chgo-odoo (approved):
Thanks again for your work on this @bso-odoo 🙏 

Had a brief look at the changes you made in the XML and it looks okay at first sight. Thanks for fixing the classes xpath that were not using any separator 😅

qsm-odoo:
> you can't rebase-ff.

Weird?
@robodoo rebase-ff

bso-odoo:
> Thanks for fixing the classes xpath that were not using any separator 😅

@chgo-odoo Note that without value, the default separator is `,` - which leads to very stange values `class="some_class other_class,added_class"` - making both last classes unrecognized :grimacing:

chgo-odoo:
> @chgo-odoo Note that without value, the default separator is `,` - which leads to very stange values `class="some_class other_class,added_class"` - making both last classes unrecognized 😬

Yep, saw that ! Weird that we manage to introduce this multiple times, I guess we really need to triple check on our side. I shared the communication one more time with the team so we should be good now 😅

sobo-odoo:
@robodoo delegate=agau-odoo

theme_yes/views/snippets/s_quotes_carousel.xml
agau-odoo, line 1: I would mention the community PR's commit in the commit message to clarify why you're making these modifications.

test_themes/tests/test_new_page_templates.py
```diff
@@ -190,6 +190,13 @@ def check(theme_name, website):
                                 conflict.difference_update(white_list)
                                 if len(conflict) > 1:
                                     errors.append("Using %r, view %r contains conflicting classes: %r in %r (according to pattern %r)" % (theme_name, view.key, conflict, classes, conflicting_classes_re.pattern))
+                        for el in html_tree.xpath('//*[@style]'):
+                            styles = el.attrib['style'].split(';')
+                            properties = list(map(lambda x: x.split(':')[0].strip(),
+                                filter(lambda x: x, styles)
+                            ))
```
agau-odoo, line 197: This is very hard to read
Maybe we should extract some variables?
I also like to name the loop variable with a clear name. In this case, I would rename `x` with `style`
`filter(lambda x: x, styles)` is a little tricky to understand, maybe worth a comment?

agau-odoo: I would also consider naming this variable `property_names` instead of `properties`, but your choice on this

test_themes/tests/test_new_page_templates.py
```diff
@@ -190,6 +190,13 @@ def check(theme_name, website):
...
+                            if len(properties) != len(set(properties)):
+                                errors.append("Using %r, view %r contains duplicate style properties: %r" % (theme_name, view.key, el.attrib['style']))
```
agau-odoo, line 199: It could be useful to hint at which property is duplicated, would that be easy to do?

test_themes/tests/test_new_page_templates.py
```diff
@@ -210,3 +217,41 @@ def check(theme_name, website):
...
+            ])
+            for view in views:
+                try:
+                    xml_tree = etree.fromstring(view.arch_db)
+                except etree.LxmlError:
+                    _logger.error("Using %r, view %r cannot be parsed: %r", module_name, view.key, view.arch_db)
+                    errors.append("Using %r, view %r cannot be parsed: %r" % (module_name, view.key, view.arch_db))
+                    continue
+                for el in xml_tree.xpath('//attribute[@add] | //attribute[@remove]'):
+                    name = el.attrib['name']
```
agau-odoo, line 247:
```suggestion
                    attribute_name = el.attrib['name']
```

agau-odoo (changes requested):
Here are some more comments:

In commit 1, some themes are missing from the module list. Please check all your commits for this.

test_themes/tests/test_new_page_templates.py
bso-odoo: I did that initially, but it was actually more helpful when fixing issues to see the actual values, to know what was supposed to have been removed. And since it is not supposed to be drawn, I chose to keep it simple.

test_themes/tests/test_new_page_templates.py
bso-odoo: Added an intermediary variable `non_empty_styles` instead of a comment.

theme_yes/views/snippets/s_quotes_carousel.xml
bso-odoo: Referenced this PR from the community commit instead. (Note that I cannot reference a specific commit before it is merged)

agau-odoo (approved):
LGTM! @robodoo r+

qsm-odoo:
image

@agau-odoo robodoo is only listening to you if you put the command on its own line in your message (line starting by `@robodoo` then only commands). Can you r+ again?

agau-odoo:
@robodoo r+

agau-odoo:
@robodoo r-

agau-odoo (approved):
@robodoo r+

PR #935, opened 2024-09-25
[IMP] theme_nano: revamp the theme

chgo-odoo:
This commit revamps the design of the `theme_nano`.

task-4178081
part of task-4177975

- requires https://github.com/odoo/odoo/pull/181369


theme_nano/views/snippets/s_parallax.xml
```diff
@@ -4,20 +4,11 @@
 <template id="s_parallax" inherit_id="website.s_parallax">
     <!-- Section -->
     <xpath expr="//section" position="attributes">
-        <attribute name="class" add="pt48 pb48" separator=" "/>
-        <attribute name="data-oe-shape-data">{"shape":"web_editor/Rainy/10", "flip":["x","y"]}</attribute>
+        <attribute name="class" add="pt48 pb48 o_not_editable" separator=" "/>
```
robinlej, line 7: Why do we need `.o_not_editable` here? That looks like something we'd want on the base snippet if it's needed?

theme_nano/views/snippets/s_references.xml
```diff
@@ -6,6 +6,13 @@
     <xpath expr="//section" position="attributes">
         <attribute name="class" add="o_cc5" remove="o_cc1" separator=" "/>
     </xpath>
+    <xpath expr="//div[hasclass('col-lg-12')]" position="replace" mode="inner">
+        <h2 style="text-align: center;">Trusted by industry leaders</h2>
+    </xpath>
+    <xpath expr="//div[hasclass('col-lg-2')]" position="attributes">
+        <attribute name="class" add="offset-lg-1" separator=" "/>
+    </xpath>
+    <xpath expr="(//div[hasclass('col-lg-2')])[6]" position="replace" mode="inner"/>
```
robinlej, line 15: Why `mode="inner"`? It should be plainly removed, shouldn't it?

robinlej (changes requested):
Do you take those kind of things into account? I ask particularly because you specified that "the color palette has been adjusted to a `high-contrast` style", but it fails in such cases.

image

And the color definitely has to be adapted on the s_big_number:
image

I also have a constrast issue on the s_image_punchy, but I think it's just due to the image being overridden by the configurator, so not much you can do.

For the profile pictures: do we need 705*705px? They don't seem to be used anywhere above 350px. Maybe the 420px from the previous ones was good enough?

robinlej:
Oops, I just saw that in the end it wasn't ready for tech review. :sweat_smile:

chgo-odoo:
It's fine, I'll fix your comments within my next push 😄 Most of the design should be validated, at least I hope 🤞

theme_nano/views/snippets/s_parallax.xml
chgo-odoo: I think you're right and we should remove this class. By default the `s_parallax` is meant to receive other snippets within its DOM but in this specific scenario I wanted to use it as a "banner" without being editable. But the more I think about it the more I think you're right, it doesn't make sense to force the class here if it's not the case on the default snippet, and restricting the edition makes no sense as it's something the user might want to do.

theme_nano/views/snippets/s_key_images.xml
```diff
@@ -2,23 +2,62 @@
...
-        Customized solutions for your business needs
+    <xpath expr="//p[hasclass('lead')]" position="replace">
+        <p class="lead" style="text-align: center;">Some of the latest projects we had the pleasure to work on</p>
     </xpath>
-    <xpath expr="(//div[hasclass('col-lg-3')])[2]//p[3]" position="replace" mode="inner">
-        Expert advice to navigate complex challenges
+
+    <xpath expr="//div[hasclass('col-lg-3')]" position="attributes">
+        <attribute name="class" add="col-lg-6 o_animate o_animated o_anim_from_left o_anim_slide_in" remove="col-lg-3" separator=" "/>
+        <attribute name="style" add="--wanim-intensity: 3; animation-name: o_anim_slide_in_left; animation-play-state: running;" separator="; "/>
```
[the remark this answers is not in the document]
chgo-odoo: Didn't know about this class, I opened a CSS fix here: https://github.com/odoo/odoo/pull/181087

Do you think you solution is more adapted? If yes I'll close my PR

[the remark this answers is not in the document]
robinlej: Oh no, if your one-line CSS fix doesn't introduce unwanted side-effects, that is definitely a better choice. :smile:

chgo-odoo:
Thanks @robinlej, I've adapted the PR with the requested changes, could you have a look at it please?

bso-odoo:
Weird, this preview does not work... it's most likely a bug in the preview mode, what do you think ? @bvr-odoo 
image

chgo-odoo:
We didn't have the use case back when the modal was implemented and it's true that this snippet is one of the first animated so I created [a pull request](https://github.com/odoo/odoo/pull/181087) for this scenario, but it's nice if BVR can check this PR as he knows well the snippet modal environment

bso-odoo:
Not originating in this PR, but `s_quotes_carousel` has the same text on all slides. - Not sure this is to be fixed here.

chgo-odoo:
Nice catch, should be fixed in another PR, I'll create a task in our pipeline

bso-odoo:
`nano_screenshot.jpg` and `theme_nano.svg` are out of sync.

chgo-odoo:
Screenshot wasn't up-to-date, thanks 😅

bso-odoo:
`congifugurator.s_cover` is still defined while it has been removed from the manifest (there might be others)

chgo-odoo:
Seems to be the only one affected, `s_images_wall` is the only other one that is no more used in the theme

theme_nano/views/new_page_template.xml
```diff
@@ -193,7 +184,13 @@
 
 <template id="new_page_template_about_personal_s_numbers" inherit_id="website.new_page_template_about_personal_s_numbers">
     <xpath expr="//section" position="attributes">
-        <attribute name="class" add="o_cc2" remove="o_cc3" separator=" "/>
+        <attribute name="class" remove="o_cc3" separator=" "/>
```
bso-odoo, line 187: Doesn't doing this leave a remaining `o_cc` without `o_ccN` ? (several times in this commit - and, I never thought about checking that in the previous themes - maybe to be handled in a cleanup for already merged themes later)

theme_nano/views/snippets/s_big_number.xml
```diff
@@ -4,13 +4,13 @@
...
     <xpath expr="//section" position="attributes">
-        <attribute name="data-oe-shape-data">{"shape":"web_editor/Bold/06_001"}</attribute>
+        <attribute name="data-oe-shape-data">{"shape":"web_editor/Bold/06_001",'colors':'c3':'o-color-5'}</attribute>
         <attribute name="class" add="o_cc4" remove="o_cc5" separator=" "/>
     </xpath>
 
     <!-- Shape -->
     <xpath expr="//div[hasclass('o_we_shape')]" position="replace">
-        <div class="o_we_shape o_web_editor_Bold_06_001"/>
+        <div class="o_we_shape o_web_editor_Bold_06_001" style="background-image: url('/web_editor/shape/web_editor%2FBold%2F06_001.svg?c3=o-color-5&amp;);"/>
```
bso-odoo, line 13: Single quote is not closed, and ending with `&amp;` seems a bit weird

theme_nano/views/new_page_template.xml
chgo-odoo: This one is a tricky one, I do agree 😅 

I do agree that with that `add=""` and `remove=""` it should be empty and thus `o_cc` but it seems to keep the value from the default template so `o_cc5`.

I made a previous try by adding `o_cc5` but runbot told me there were duplicated classes. By checking the new pages in the modal, it seems fine, but these templates should be refined later, indeed

bso-odoo:
Fixed the following in `s_big_number`:
- Missing object around colors in shape definition.
- Extra semi-colon at end of URL.

Also rebased.

chgo-odoo:
Thanks for the fix @bso-odoo, fixed two missing snippets in the `tour.js`

bso-odoo:
@robodoo r+

bvr-odoo:
> Weird, this preview does not work... it's most likely a bug in the preview mode, what do you think ? @bvr-odoo image Maybe we should handle the "animated blocks" and temporarily make them visible during previews ? (Maybe injecting CSS about that in the preview would work)

The `o_animate` elements are hidden before their widget starts.
So indeed, we need to handle this case in the modal :+1:

PR #937, opened 2024-09-25
[IMP] theme_yes: revamp theme

chgo-odoo:
| Master | This PR |
|--------|--------|
| image | image | 

task-4178092
part of task-4177975


theme_yes/views/snippets/s_image_text_overlap.xml
```diff
@@ -0,0 +1,24 @@
...
+
+<template id="s_image_text_overlap" inherit_id="website.s_image_text_overlap">
+  <xpath expr="//section" position="attributes">
+    <attribute name="class" add="o_cc o_cc2" separator=" "/>
+  </xpath>
+  <xpath expr="//div[hasclass('row')]" position="attributes">
+    <attribute name="data-row-count" add="10" remove="9" separator=" "/>
+  </xpath>
+  <xpath expr="(//div[hasclass('o_grid_item')])[1]" position="attributes">
+    <attribute name="style" add="grid-area: 1 / 4 / 11 / 13; --grid-item-padding-x: 0px;" remove="grid-area: 1 / 1 / 10 / 8;" separator=";"/>
```
bso-odoo, line 12: Spotted by not-yet merged test:
```
Using 'Yes Theme', view 'website.configurator_s_image_text_overlap' does not specify 'g-height-10' for grid item '1 / 4 / 11 / 13' (['o_grid_item', 'o_grid_item_image', 'order-lg-0', 'g-col-lg-7', 'g-height-9', 'col-lg-7'])
```

chgo-odoo:
Thanks @bso-odoo `g-height` were wrong as well, should we good now

bso-odoo:
When adding a new page > About > Second column, second template: is that wide green area on the bottom of the block intended (block is set to full height) ? (same thing also appears in Basic column 2's 2nd template - but there it's the only block, so it's not that weird)
image

chgo-odoo:
I do agree that this isn't great, but it's not related to this PR. I think we're planning to review these templates. I'd rather let it as is or fix it in another PR

bso-odoo:
@robodoo r+

bso-odoo:
@robodoo r+

PR #942, opened 2024-09-26
[FIX] theme_bistro, theme_beauty: adapt tours

Brieuc-brd:
Following the reorganization of price list snippets, this commit adapts the tour of themes that currently use them.

task-4212916

Requires:
- https://github.com/odoo/odoo/pull/181510


qsm-odoo (approved):
@robodoo r+

PR #968, opened 2024-09-29
[FIX] theme_(yes, loftspace): align themes preview

chgo-odoo:
This commit fixes an issue about these two themes previews not being similar to the other ones.

Prior to this commit, some difference were existing between themes screenshot. With the revamp of the themes, these differences were fixed. But an other issue was introduced within the screenshots of the themes I worked on. Other modules were installed, making the header filled with more `nav-link`. Sometimes the editor button was even visible.

This commit replaces these screenshots with new ones, aligned on the other themes.

task-4218255


qsm-odoo:
Being moved to https://github.com/odoo/design-themes/pull/969
Don't push here or there anymore, if you need to apply fixes, create a new PR to merge once the main one is merged.

PR #972, opened 2024-09-30
[FIX] theme_bookstore: fix multiple issues in the theme

anso-odoo:
This commit fixes multiples issues in the theme:

- Wording
- Shapes that should appear only in the configurator
- Useless xpaths

task-4221731


sobo-odoo:
@anso-odoo FYI I force-pushed to:
- remove unused images from the manifest => tell me if I am wrong but the `images_theme_preview` images are only used to be called in the `href` right ? Since it was not the case for the banner one, I removed it, and I removed the duplicated image_text one :slightly_smiling_face: 
- fix the shape of `s_title` => you forgot to remove the attribute from the "normal" snippet and you set the wrong one for the "configurator" one :wink: 

=> see [diff](https://github.com/odoo/design-themes/compare/327b189260f2531d849d570c4eccfe44c993467f..df6362eaab59d6ccfe560548cb53a53bbe417b1f)

Note, I think that s_image_text_default_image is used too many times (3 identical images in the same pages seems too much) :grimacing: But we can fix it later when we will fix the svg and the fallback images :slightly_smiling_face:

sobo-odoo:
@robodoo r-

sobo-odoo (approved):
@robodoo r+

PR #976, opened 2024-10-01
[FIX] test_themes, theme_*: fix not working theme tours

sobo-odoo:
*: loftspace, real_estate

In commit [1], the test for the "Loftspace" and "Real Estate" themes has been disabled, as their tour was not working.

This commit fixes the tours and reenables the test for these themes. The issue was coming from the fact that they were trying to drop more snippets than there really are, which made an acces to the array at an inexisting index.

[1]: https://github.com/odoo/design-themes/commit/04f90cd883826bb8e4cbf8975aff6c14bf3c64db


qsm-odoo (approved):
@robodoo r+

PR #982, opened 2024-10-08
[FIX] test_themes: fix test_01_theme_upgrade_post_copy

bvr-odoo:
Since the revamp of the "Nano" theme with this commit [1], the "option_footer_scrolltop" template is no longer enabled for this theme. This causes the "test_01_theme_upgrade_post_copy" test to fail, as it relies on this template being enabled.

[1]: https://github.com/odoo/design-themes/commit/47568ec66fac4714c53553d021b3f1d7bd252dea


bso-odoo:
@robodoo r+

PR #984, opened 2024-10-09
[FIX] theme_cobalt, theme_paptic: add some missing image classes

sobo-odoo:
In some theme customizations, some replaced images do not have the usual image classes `img img-fluid mx-auto` anymore. The `img-fluid` one is needed for grid images in mobile view, as they are overflowing without it.

This commit adds these missing classes on the identified problematic images.


qsm-odoo (approved):
@robodoo r+

I'll handle the forward-port

PR #987, opened 2024-10-14
[FIX] theme_artists: add missing `o_colored_level` class

chgo-odoo:
This PR adds a missing `o_colored_level` class on the section.

With the themes revamp, we reviewed the design of a lot of snippets, which sometimes included colors adjustments using `o_cc o_cc*` classes.

While the visual result is okay on the page, the title of the snippet will be invisible in the modal as two `o_cc o_cc*` are nested without having an `o_colored_level` class on the section.

| Master | This PR |
|--------|--------|
| image | image | 

task-4255088
related to task-4177975


chgo-odoo:
@robodoo help

chgo-odoo:
To be fixed at the snippet level instead of theme level.

PR #991, opened 2024-10-17
[IMP] test_themes: fix test for `s_rating` inner snippet in `s_reviews_wall`

chgo-odoo:
This commit adds a line within the `new_page_templates` test in order to allow the snippet `s_rating` to be used with `s_rating_no_title` option activated. Since this inner snippet was never used within another one before, we didn't need to add such a line.

- requires https://github.com/odoo/odoo/pull/183965

task-4264114
part of task-4077427


chgo-odoo:
Probably not needed

PR #1000, opened 2024-10-28
[IMP] theme_*: adapt `s_numbers_framed` across design themes

chgo-odoo:
*: artists, avantgarde, aviato, beauty, bewise, bistro, bookstore, buzzy, clean, cobalt, enark, graphene, kea, kiddo, loftspace, monglia, nano, botes, odoo_experts, orchid, paptic, real_estate, treehouse, vehicle, yes,zap.

This PR adapts the design of the `s_numbers_framed` occurrences across design themes.

- requires https://github.com/odoo/odoo/pull/175700

task-4094384
Part of task-4077427

### DIFF
--- a/theme_beauty/views/snippets/s_numbers_framed.xml
+++ b/theme_beauty/views/snippets/s_numbers_framed.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<template id="s_numbers_framed" inherit_id="website.s_numbers_framed">
+    <!-- Section -->
+    <xpath expr="//section" position="attributes">
+        <attribute name="class" add="o_cc5" separator=" "/>
+    </xpath>
+    <!-- KPI's -->
+    <xpath expr="//div[hasclass('col-lg-4')][1]" position="attributes">
+        <attribute name="class" add="o_cc3" remove="o_cc2" separator=" "/>
+    </xpath>
+    <xpath expr="//div[hasclass('col-lg-4')][2]" position="attributes">
+        <attribute name="class" add="o_cc3" remove="o_cc2" separator=" "/>
+    </xpath>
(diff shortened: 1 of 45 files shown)


theme_beauty/views/snippets/s_numbers_framed.xml
chrisdt1998, line 1: There's something weird happening with this theme. When you select the block, it has white numbers but then it flickers to black once it's been chosen.
image
image

theme_bewise/views/customizations.xml
chrisdt1998, line 1: Same thing here
image
image

theme_loftspace/views/snippets/s_numbers_framed.xml
chrisdt1998, line 1: Same here:
image

image

theme_monglia/views/customizations.xml
chrisdt1998, line 1: Same here:
image
image

theme_notes/views/snippets/s_numbers_framed.xml
chrisdt1998, line 1: Same here:
image
image

chrisdt1998 (commented):
There are some issues with some of the blocks. Do you know why this happens?

chgo-odoo:
Thanks @chrisdt1998 👍 

It's due to a missing `o_colored_level` class on the section which is mandatory when we nest `o_cc` (which we do in some themes). It should be fixed now

chrisdt1998:
@chgo-odoo Ah I see. However, I think it would make more sense to instead add o_colored_level to the base snippet section element instead of doing it for each theme. Also, the reason for the 'flicker' is because  
image
adds the o_colored_level class to the section element. That's why in the default snippet, even though you don't explicitly add the o_colored_level class, it will be added there when you drag and drop the snippet.

TLDR: please add o_colored_level to the section element of the default snippet and adjust appropriately for the theme snippets.

PR #1005, opened 2024-11-05
[IMP] theme_*: adapt `s_newsletter_box` to themes

mano-odoo:
*: theme_artists, theme_nano

This commit adapts the new `s_newsletter_box` snippet to themes.

task-4306880
Part of task-4077427


mano-odoo:
Closedd because can't adapt to themes as the template could be unreachable if `website_mass_mailing` is not installed

PR #1009, opened 2024-11-12
[IMP] core: tools.config.is_testing flag

kmagusiak:
Flag to check whether Odoo was started in test mode.

task-4270485
odoo/odoo#186924


Xavier-Do:
@robodoo r+

PR #1011, opened 2024-11-15
[FIX] test_themes: allow proper patch website_swit

Williambraecky:
..cher

Defines dataset directly on website elements such that it can be overridden by other modules if necessary.
This was done due to the override in the `test_themes` module completely overriding all attributes of the website_switcher's dropdown items.

Runbot Error 106501


Williambraecky:
robodoo r+

Williambraecky:
@Xavier-Do Could you r+, the fix remove the bad override in this module.

Williambraecky:
robodoo r+

PR #1012, opened 2024-11-18
[FIX] theme_*: fix numbers snippets layout across themes

chgo-odoo:
*: nano, notes, oddoo_experts, treehouse, zap

This commit reviews the occurrences of the `s_numbers_*` snippets to align them with the new layout of these snippets.

Prior to this commit, the layout of those snippet was "broken" due to weird structure like nested rows, .. which prevents the editor to apply a change on the whole content.

As we reworked the layout of these snippets to fix the issue, we need to apply these changes in themes as well.

task-4223179


qsm-odoo (approved):
@robodoo r+

PR #1022, opened 2024-12-04
[FIX] avantgarde + masonry block

qsm-odoo:


qsm-odoo (commented):
@robodoo rebase-ff r+

PR #1047, opened 2025-02-12
[FIX] theme_cobalt, *: update xpath to removed s_nb_column_fixed

dery-odoo:
*: theme_anelusia, theme_monglia, theme_nano, theme_notes, theme_odoo_experts, theme_orchid, theme_paptic,theme_yes, theme_zap

Multiple snippet template where changed to ensure that the title is not considered as a column and allowing changing the nb of column again. This commit adapt all the xpath that were impacted by those changes.

task-4310468


qsm-odoo (approved):
@robodoo r+

PR #1061, opened 2025-05-07
[IMP] theme_artists, *: adapt `s_ecomm_cols_infos_h`

Brieuc-brd:
*: theme_loftspace, theme_nano, theme_orchid

This commit adapts customizations for `s_ecomm_cols_infos_h`.

Requires:
- https://github.com/odoo/odoo/pull/208764

task-4775848
Part of task-4252024


test_themes/tests/test_new_page_templates.py
```diff
@@ -110,7 +110,7 @@
     's_carousel_indicators_dots', 's_quotes_carousel', 's_rating_no_title',
     's_carousel_indicators_hidden', 's_blog_post_big_picture',
     's_newsletter_list', 's_event_upcoming_snippet', 's_event_event_picture',
-    's_newsletter_subscribe_form',
+    's_newsletter_subscribe_form', 's_attributes_img', 's_attributes_col',
```
qsm-odoo, line 113: Adding items to this list means you are doing something wrong: indeed, I missed it in my community review but they should be called `s_attributes_horizontal_img` and `s_attributes_horizontal_col`. Or don't add the word "horizontal" in your snippet name in the first place, as you wish :shrug:

qsm-odoo (changes requested):

test_themes/tests/test_new_page_templates.py
Brieuc-brd: As we will also introduce another snippet called `s_attributes_vertical`, I'd prefer to keep the word "horizontal" 🙂  I've made the changes, could you take a look at it?
Thanks in advance!

PR #1062, opened 2025-05-07
[FIX] theme_test_custo: prevent failing test_website_theme_preview tour

hubvd:
When running the test test_website_theme_preview, we have the following error:

`tour.steps has to be a function that returns TourStep[]`


qsm-odoo (approved):
@robodoo r+

PR #1078, opened 2025-05-15
[FIX] theme_test_custo: fix nightly theme preview tour

qsm-odoo:
The theme preview feature before selection was disabled in [1]. This commit disables the related tour steps that were testing it a nightly tour... that was red since then.

[1]: https://github.com/odoo/odoo/commit/7cb71e9479df0ee9af0b7ad39302857666726177

Related to task-3454790


qsm-odoo (commented):
@robodoo r+

PR #1102, opened 2025-06-18
[FIX] theme_test_custo: correct tour trigger to look for svg images

hubvd:
Since [1], the test "test_01_menu_hierarchies" fails.
Note that the check is not as strong as before as we cannot check the
source of the image.
Since [2], the btn-primary is now btn-secondary

---

[1]: https://github.com/odoo/odoo/commit/3cc5ff6e96493f2b068090f880987def3787895e
[2]: https://github.com/odoo/odoo/commit/eca4d1b1d4ea58959793b0a2cbfa443308759372
runbot-226641

### DIFF
--- a/theme_test_custo/static/tests/tours/theme_menu_hierarchies.js
+++ b/theme_test_custo/static/tests/tours/theme_menu_hierarchies.js
@@ -39,7 +39,7 @@ wTourUtils.registerWebsitePreviewTour('theme_menu_hierarchies', {
         run: "click",
     }, {
         content: 'The theme custom footer template should be listed and selected',
-        trigger: '[data-container-title="Footer"] [data-label="Template"] button.btn-primary img[src*="/theme_test_custo"]',
+        trigger: '[data-container-title="Footer"] [data-label="Template"] button.btn-secondary svg.theme_test_custo_footer',
         run: () => null, // It's a check.
     }, {
         content: 'Click on header',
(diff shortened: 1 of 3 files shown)


hubvd:
@robinlej @loco-odoo What do you think ? Maybe the Img component should have an attribute with the source to be able to check it's value ?

```xml
<svg t-att-data-src="props.src" ... />
```

theme_test_custo/static/tests/tours/theme_menu_hierarchies.js
```diff
@@ -39,15 +39,15 @@ wTourUtils.registerWebsitePreviewTour('theme_menu_hierarchies', {
         run: "click",
     }, {
         content: 'The theme custom footer template should be listed and selected',
-        trigger: '[data-container-title="Footer"] [data-label="Template"] button.btn-primary img[src*="/theme_test_custo"]',
+        trigger: '[data-container-title="Footer"] [data-label="Template"] button.btn-primary svg',
```
robinlej, line 42: Hi @hubvd, this won't work. (I mean, it will work, but it won't test that the custom template is selected, this selector could match any template.)

I don't think that we need an actual data-src everywhere either. This could change in the future, but for now I think it makes more sense to work on an case-by-case basis.

What I'd suggest is to add a specific class on the builder option, and target it through the class in the test:
```xml
<Img src="'/theme_test_custo/static/src/img/template_header_opt.svg'" class="'theme_test_custo_header'"/>
```

hubvd:
I rebased an turns out another change has been made that breaks the tour https://github.com/odoo/odoo/pull/213019/commits/eca4d1b1d4ea58959793b0a2cbfa443308759372#diff-d3ccab48ea391b9812d735e007298e1d86e6b93b8848fb8204b5be6a6524c40a

bso-odoo:
@robodoo delegate=robinlej

robinlej (approved):
LGTM :+1: 
Should I r+ @hubvd ?

hubvd:
Yes, thanks :)

hubvd:
Sure

PR #1105, opened 2025-06-25
[IMP] theme_buzzy, *: fix header hierarchy for better SEO

SergeBayet:
*: theme_nano, theme_real_estate

Refactored design-themes modules templates to use a consistent and semantically correct heading structure (h1, h2 etc.). This improves SEO.

task-4680875


sobo-odoo:
@robodoo delegate=bvr-odoo

PR #1106, opened 2025-06-25
[FIX] theme_test_custo: correct tour trigger to look for svg images

hubvd:
Since [1], the test "test_01_menu_hierarchies" fails.

Note that the check is not as strong as before as we cannot check the source of the image.

Since [2], the btn-primary is now btn-secondary

---

[1]: https://github.com/odoo/odoo/commit/3cc5ff6e96493f2b068090f880987def3787895e
[2]: https://github.com/odoo/odoo/commit/eca4d1b1d4ea58959793b0a2cbfa443308759372

runbot-226641


hubvd:
Backport of https://github.com/odoo/design-themes/pull/1102

bso-odoo:
@robodoo delegate=robinlej

robinlej:
@robodoo r+

PR #1108, opened 2025-07-03
[IMP] theme_test_custo: update js import

thjo-odoo:
This commit updates the js import for ImageShapeOptionPlugin as it was moved from website to html_builder in: odoo/odoo#215136.

task-4247642


tde-banana-odoo:
@robodoo please die

PR #1111, opened 2025-07-08
[REM] theme_test_custo: no need waitIframeIsReady

pipu-odoo:
In this commit, we remove waitIframeIsReady. This is not usefull until
it is checked in tour_step_automatic.
We also remove empty run.


rfr-odoo:
@robodoo help

PR #1114, opened 2025-07-11
[IMP] test_themes, theme_*: adapts themes for new parallax structure

bvr-odoo:
This commit adapts the themes to the new Parallax effect structure. We now wrap the parallax element in another element with "overflow: hidden" to avoid setting "overflow: hidden" on sections. That was causing elements like the dropdown of the search bar to be cut off and not fully visible.

task-4926420


qsm-odoo (changes requested):
- Commit message nitpicking: adapt instead of adapts
- Also rebase both PR one last time while you are at it, then I'll test and merge this afternoon :pray:

PR #1126, opened 2025-08-04
[FIX] theme_*: adapt xpath for s_timeline(_images, _list) changes

xlu-odoo:
*: theme_artists, theme_buzzy, theme_nano

Before, the timeline (*) dots were invisible when using preset 4 because they shared the same color as the background (`text-o-color-1` on `o-color-1` background).

The fix in the relative PR implements a solution that ensures dots are always visible by using button primary colors for all presets.

Thus, this commit adapts xpath to ensure compatibility with the changes.

requires: https://github.com/odoo/odoo/pull/221505

task-4922302


mano-odoo (approved):
Good job, LGTM ✅

xlu-odoo:
> @xlu-odoo I am guessing you still have to adapt this one?

Yep, done ✅

qsm-odoo (approved):
@robodoo r+

PR #1128, opened 2025-08-11
[IMP] website, website_sale: improve header options

romo-odoo:
Update a template name to match the new name.

See also : odoo/odoo#221703

task-4556047


bso-odoo:
@robodoo delegate=SergeBayet

PR #1131, opened 2025-08-19
[IMP] website, html_builder: move image and background options

abd-msyukyu-odoo:
As mass_mailing will soon use the `html_builder` instead of the legacy editor, some useful options currently exclusive to `website` can be moved to be globally available.

task-4247642


sobo-odoo:
@robodoo delegate=FrancoisGe

PR #1132, opened 2025-08-20
[FIX] theme_*: fix duplicate palette crashing render in theme tab

duau-odoo:
*: theme_anelusia, theme_avantgarde, theme_buzzy, theme_real_estate, theme_treehouse

Since website refactor [1],a silent error would occur while rendering the theme tab, because of the use of `t-foreach`, leading to nothing being displayed depending of the theme.

This was caused by a duplicate key in the
`$o-selected-color-palettes-names` scss variable, added by some theme, which would crash the `t-foreach` template loop in `website.ThemeColorsOption`.

Steps to reproduce:
- Create a website using any of the above themes, through the wizard
- Enter edit mode
- Click on theme tab
- Nothing is rendered

[1]: https://github.com/odoo/odoo/commit/9fe45e2b7ddbbfd0445ffe25a859e67a316d02b2


qsm-odoo (changes requested):

qsm-odoo (commented):
Testing a bit, that's funny: when they removed the "saveAs" parameter to override the homepage tour instead of fully redefining it... that just broke our logs :grimacing: 

Now you have ~30:
```
╔═════════════════════════╗
║ TOUR homepage SUCCEEDED ║
╚═════════════════════════╝
```
So, if all pass that's fine... but if one fails, you don't clearly see which theme broke :grimacing: Not critical, but we should do something about it... I'll add it in the follow-up task.

PR #1149, opened 2025-09-15
[IMP] test_themes, *: remove obsolete .btn-block

ldau-odoo:
*: theme_vehicle

This PR removes the `btn-block` utility Bootstrap class because it's obsolete since Bootstrap 5.

task-4637091


test_themes/tests/test_new_page_templates.py
```diff
@@ -13,7 +13,7 @@
 
 CONFLICTUAL_CLASSES = [
     ['btn-outline-primary', 'btn-primary', 'btn-secondary'],
-    ['btn-block', 'btn-outline-primary'],
```
chgo-odoo, line 16: I think you can remove this entire line since there is only one class name left. It's probably not useful at all 😉

qsm-odoo (approved):
This theme part looks good:

@robodoo r+

PR #1151, opened 2025-09-25
[IMP] website: Introduce high-level cache for website pages

Gorash:
ormcache invalidation in test_02_homepage_tour_every_theme test.

The configurator_apply method and various methods used for theme added
on `ir.module.module` from website write directely on `ir.model.data`
and update attachments, views, xmlids. They are missing invalidations
of template ormcache and others.

Adding a high-level cache revealed the problem, but this issue
already exists, it impacts all template caches such as `website.page`
and `ir.qweb`.

see https://github.com/odoo/odoo/pull/224487
see https://github.com/odoo/odoo/pull/88276

### DIFF
--- a/test_themes/tests/test_crawl.py
+++ b/test_themes/tests/test_crawl.py
@@ -49,4 +49,11 @@ class Crawler(HttpCase):
         Website = self.env['website']
         websites_themes = Website.get_test_themes_websites()
         for website in websites_themes:
+            # TODO: remove this invalidation and invalidation in theme feature.
+            # They are missing invalidations of template ormcache and others.
+            # The configurator_apply method and various methods used for theme
+            # added on `ir.module.module` from website write directly on
+            # `ir.model.data` and update attachments, views, xmlids.
+            self.env.registry.clear_cache('templates')
+
             self.start_tour(f"/web?fw={website.id}", 'homepage', login='admin')


test_themes/tests/test_crawl.py
```diff
@@ -49,4 +49,11 @@ def test_02_homepage_tour_every_theme(self):
         Website = self.env['website']
         websites_themes = Website.get_test_themes_websites()
         for website in websites_themes:
+            # TODO: remove this invalidation and invalidation in theme feature.
+            # They are missing invalidations of template ormcache and others.
+            # The configurator_apply method and various methods used for theme
+            # added on `ir.module.module` from website write directely on
+            # `ir.model.data` and update attachments, views, xmlids.
+            self.env.registry.clear_cache('templates')
```
qsm-odoo, line 57: TBH I am not sure to understand: how is what you describe related to starting tours? Is not there a bug/something to do in the community code? 🤔

test_themes/tests/test_crawl.py
```diff
@@ -49,4 +49,11 @@ def test_02_homepage_tour_every_theme(self):
         Website = self.env['website']
         websites_themes = Website.get_test_themes_websites()
         for website in websites_themes:
+            # TODO: remove this invalidation and invalidation in theme feature.
+            # They are missing invalidations of template ormcache and others.
+            # The configurator_apply method and various methods used for theme
+            # added on `ir.module.module` from website write directely on
```
qsm-odoo, line 55: directly

qsm-odoo (changes requested):
There are still comments to consider on the community PR IMO, but I let you judge. And I let you the r+ over here once ready.

@robodoo delegate+

test_themes/tests/test_crawl.py
Gorash: Indeed, there's a change to be made in the community theme section, but the code is very difficult to understand, and the number of places where xmlid style records are modified without invalidating the corresponding models is significant. I've asked for help understanding the system, but the fix can only be done by those who developed and modified this code (in community and themes).

The cache errors are already present. The higher-level cache reveals an invalidation bug by accumulating theme installations with template and attachment modifications.

PR #1159, opened 2025-09-29
[FIX] test_themes: run configurator standalone test under mocked request

SergeBayet:
Commit [1] copies configurator pages during the flow. When the
standalone test leaves `website_id` in the context, that copy now tries
to create generic views with a website-scoped environment and hits the
ir.ui.view guard. Running the configurator inside MockRequest mimics the
real HTTP context (request-driven website, no forced id on the ORM), so
the copy succeeds and the test passes again.

[1]: https://github.com/odoo/odoo/commit/4231c9e545fdebdad034a4418f2a11e6691bac61

runbot-232963

### DIFF
--- a/test_themes/tests/test_theme_standalone.py
+++ b/test_themes/tests/test_theme_standalone.py
@@ -1,6 +1,8 @@
 # Part of Odoo. See LICENSE file for full copyright and licensing details.
 
 from unittest.mock import patch
+
+from odoo.addons.http_routing.tests.common import MockRequest
 from odoo.tests import standalone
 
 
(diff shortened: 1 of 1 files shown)


test_themes/tests/test_theme_standalone.py
```diff
@@ -28,7 +30,7 @@ def test_02_theme_default_generate_primary_templates(env):
 
     def fake_website_api(self, path, payload):
         # Only the configurator custom_resources endpoint is used here
-        assert path.startswith("/api/website/2/configurator/custom_resources")
+        assert path.startswith(f"/api/website/2/configurator/custom_resources")
```
qsm-odoo, line 33: :thinking:

qsm-odoo (changes requested):
\+ problem in your commit message which now mentions "Commit 1" without ref

PR #1205, opened 2026-01-06
[IMP] theme_[*]: update s_cover xpath in design themes

ptpu-odoo:
*: anelusia, artists, aviato, bewise, buzzy, graphene, loftspace, nano,
   paptic, real_estate, treehouse, vehicle, yes

Following the change in s_cover's base structure (section padding moved
from `pt232/pb232` to `pt72/pb72` + a new column with `pt160/pb160`), the
xpaths in design themes were updated to apply each theme's custom
spacing on the column instead of the section.

The section padding is now consistently set to `pt72/pb72` across all
themes. The remaining spacing is distributed to the column inside as
follows:

- For snippets with the `o_three_quarter_height` class: `3/4th` of the
  original section padding is applied to the column, as the section
  height is constrained and full padding would shrink the content area.
- For all others: `72` is subtracted from the original section padding
  and the remainder is applied to the column.
    
task-[5346415](https://www.odoo.com/odoo/project/974/tasks/5346415)
Community PR: https://github.com/odoo/odoo/pull/242024

### DIFF
--- a/theme_artists/views/new_page_template.xml
+++ b/theme_artists/views/new_page_template.xml
@@ -103,7 +103,16 @@
 
 <template id="new_page_template_about_s_cover" inherit_id="website.new_page_template_about_s_cover">
     <xpath expr="//section" position="attributes">
-        <attribute name="class" add="o_three_quarter_height" remove="o_full_screen_height pt40 pb40" separator=" "/>
+        <attribute name="class" add="o_three_quarter_height" remove="o_full_screen_height" separator=" "/>
+    </xpath>
+    <xpath expr="//div[hasclass('col-lg-12')]" position="attributes">
+        <attribute name="class" add="pt56 pb32" remove="pt152 pb128" separator=" "/>
+    </xpath>
+</template>
+
+<template id="new_page_template_landing_4_s_cover" inherit_id="website.new_page_template_landing_4_s_cover">
+    <xpath expr="//div[hasclass('col-lg-12')]" position="attributes">
+        <attribute name="class" add="pt56 pb32" remove="pt40 pb40 pt152 pb128" separator=" "/>
     </xpath>
 </template>
 
@@ -158,8 +167,8 @@
 </template>
 
 <template id="new_page_template_landing_0_s_cover" inherit_id="website.new_page_template_landing_0_s_cover">
-    <xpath expr="//section" position="attributes">
-        <attribute name="class" remove="pt224 pb200" separator=" "/>
+    <xpath expr="//div[hasclass('col-lg-12')]" position="attributes">
+        <attribute name="class" add="pt160" remove="pt184 pt152 pb128" separator=" "/>
     </xpath>
 </template>
 
--- a/theme_bewise/views/customizations.xml
+++ b/theme_bewise/views/customizations.xml
@@ -4,7 +4,7 @@
 <!-- ======== COVER ======== -->
 <template id="s_cover" inherit_id="website.s_cover" name="Be Wise s_cover">
     <xpath expr="//section" position="attributes">
-        <attribute name="class" add="pt200 pb200 oe_img_bg o_bg_img_center" remove="bg-black-50 s_parallax_is_fixed parallax pt232 pb232" separator=" "/>
+        <attribute name="class" add="oe_img_bg o_bg_img_center" remove="parallax s_parallax_is_fixed bg-black-50" separator=" "/>
         <attribute name="style">background-image: url('/web/image/website.s_cover_default_image'); background-position: 50% 50%;</attribute>
         <attribute name="data-oe-shape-data">{"shape":"html_builder/Rainy/08_001","flip":[]}</attribute>
         <attribute name="data-scroll-background-ratio">0</attribute>
@@ -15,6 +15,10 @@
     </xpath>
     <!-- Remove the background image for parallax -->
     <xpath expr="//span[hasclass('s_parallax_bg_wrap')]" position="replace"/>
+    <!-- Column -->
+    <xpath expr="//div[hasclass('col-lg-12')]" position="attributes">
+        <attribute name="class" add="pt128 pb128" remove="pt160 pb160" separator=" "/>
+    </xpath>
     <!-- Title -->
     <xpath expr="//h1" position="replace" mode="inner">
         Always at the top
--- a/theme_graphene/views/customizations.xml
+++ b/theme_graphene/views/customizations.xml
@@ -4,41 +4,39 @@
 <!-- ======== COVER ======== -->
 <template id="s_cover" inherit_id="website.s_cover" name="Graphene s_cover">
     <xpath expr="//section" position="attributes">
-        <attribute name="class" add="pt160 pb160" remove="pt232 pb232 s_parallax_is_fixed" separator=" "/>
+        <attribute name="class" remove="s_parallax_is_fixed" separator=" "/>
         <attribute name="data-scroll-background-ratio">0</attribute>
         <attribute name="style">background-image: url('/web/image/website.s_cover_default_image');</attribute>
     </xpath>
     <xpath expr="//span[hasclass('s_parallax_bg_wrap')]" position="replace"/>
-    <xpath expr="//h1" position="replace"/>
-    <xpath expr="//p" position="replace"/>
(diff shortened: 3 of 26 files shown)


sobo-odoo (changes requested):
Some comments :slightly_smiling_face: 

Most are about adding a smaller padding when the 3/4 class is on the block => we don't need to set a padding that high since the block is already forced to have a height of 75% and the content is well centered :slightly_smiling_face:

theme_graphene/views/customizations.xml
```diff
@@ -4,41 +4,39 @@
...
-                    Our mission is to give customers the best experience.<br/>Extensive documentation &amp; guides, an active community,<br/>24/7 support make it a pleasure to work with us.
-                </p>
-                <p>
-                    <a t-att-href="cta_btn_href" class="btn btn-primary btn-lg o_translate_inline"><t t-out="cta_btn_text">Contact us</t></a>
-                </p>
-            </div>
-        </div>
+    <xpath expr="//h1" position="replace">
+        <h1>Making the difference.</h1>
+    </xpath>
```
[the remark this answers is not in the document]
ptpu-odoo: We went with `position="replace"` instead of `mode="inner"` because the `h1` in the `s_cover` template already has existing classes and inline styles that also need to be removed. Using `position="replace"` takes care of everything in one xpath, avoiding the need for additional xpaths to strip the classes and styles separately. 

Let me know if there's a cleaner approach I'm missing! :slightly_smiling_face:

theme_yes/views/new_page_template.xml
```diff
@@ -120,8 +120,10 @@
 
 <template id="new_page_template_about_mini_s_cover" inherit_id="website.new_page_template_about_mini_s_cover">
     <xpath expr="//section" position="attributes">
-        <!-- Defined in both theme and new page template -->
-        <attribute name="class" add="o_three_quarter_height" remove="pt40 pb40 o_three_quarter_height" separator=" "/>
+        <attribute name="class" remove="o_three_quarter_height" separator=" "/>
```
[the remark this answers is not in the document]
ptpu-odoo: > Why did you remove the class ? 

Missed that one, reverted back :+1:

theme_artists/views/new_page_template.xml
```diff
@@ -158,8 +167,8 @@
 </template>
 
 <template id="new_page_template_landing_0_s_cover" inherit_id="website.new_page_template_landing_0_s_cover">
-    <xpath expr="//section" position="attributes">
-        <attribute name="class" remove="pt224 pb200" separator=" "/>
+    <xpath expr="//div[hasclass('col-lg-12')]" position="attributes">
+        <attribute name="class" add="pt160" remove="pt184 pt152 pb128" separator=" "/>
```
ptpu-odoo, line 171: Padding top is set to `160` instead of `184` to keep the content aligned with the background line at the top, staying consistent with other `s_cover` usages.

|Before|After|
| ------------- | ------------- |
| Image | Image |

ptpu-odoo:
All design theme templates containing the `o_three_quarter_height` class have been updated to apply `3/4th` of the original section padding to the column.

Note: Switching the section height to `Auto` in these templates will result in a visible layout change. This is expected, as the padding is now applied to the column at a reduced ratio compared to the original section padding.

For templates with 100% height, no changes have been made. The column padding remains unchanged. Let me know if we should also introduce a proportional padding approach for these templates.

ptpu-odoo:
Also added vertical padding (`72` on top and bottom) to the section in the **Real Estate theme** for the following templates:
- `new_page_template_landing_2_s_cover`
- `new_page_template_gallery_s_cover`
- `new_page_template_pricing_s_cover`

| Before | After |
|----|------|
| image | image |
| image | image |
| image | image |

theme_bewise/views/customizations.xml
```diff
@@ -13,6 +13,10 @@
     <xpath expr="//div[hasclass('container')]" position="before">
         <div class="o_we_shape o_html_builder_Rainy_08_001"/>
     </xpath>
+    <!-- Column -->
+    <xpath expr="//div[hasclass('col-lg-12')]" position="attributes">
+        <attribute name="class" add="pt128 pb128" remove="pt160 pb160" separator=" "/>
+    </xpath>
```
sobo-odoo, line 21: Nitpick: move this xpath after the one for the parallax and before the one for h1 (to follow the order of the elements in the DOM)

theme_graphene/views/customizations.xml
```diff
@@ -4,41 +4,39 @@
...
-        <attribute name="class" add="pt96 pb48" remove="pt160 pb160" separator=" "/>
+        <attribute name="class" add="pt96 pb48" remove="pt72 pb72" separator=" "/>
         <attribute name="data-oe-shape-data">{"shape":"html_builder/Connections/01","colors":{"c5": "o-color-3"},"flip":[]}</attribute>
     </xpath>
     <xpath expr="//section/div[hasclass('container')]" position="before">
         <div class="o_we_shape o_html_builder_Connections_01" style="background-image: url('/html_editor/shape/html_builder/Connections/01.svg?c5=o-color-3');"/>
     </xpath>
     <xpath expr="//div[hasclass('col-lg-7')]" position="attributes">
-        <attribute name="class" add="pb56" separator=" "/>
+        <attribute name="class" add="pb56" remove="pb88" separator=" "/>
```
sobo-odoo: This one is wrong, you should also remove `pt88` to look like before :slightly_smiling_face:

theme_graphene/views/new_page_template.xml
```diff
@@ -60,6 +60,12 @@
     </xpath>
 </template>
 
+<template id="new_page_template_landing_4_s_cover" inherit_id="website.new_page_template_landing_4_s_cover">
+    <xpath expr="//div[hasclass('col-lg-7')]" position="attributes">
+        <attribute name="class" remove="pt88 pb88" separator=" "/>
+    </xpath>
+</template>
+
```
sobo-odoo: Move this lower in the file, after `new_page_template_landing_2_s_cover`, here you are in the middle of the "About" snippets :slightly_smiling_face:

theme_loftspace/views/new_page_template.xml
```diff
@@ -197,12 +197,6 @@
     </xpath>
 </template>
 
-<template id="new_page_template_landing_0_s_cover" inherit_id="website.new_page_template_landing_0_s_cover">
-    <xpath expr="//section" position="attributes">
-        <attribute name="class" remove="pt232" separator=" "/>
-    </xpath>
-</template>
-
```
sobo-odoo, line 205: Instead of removing the template, maybe decrease/remove the padding of the column, given that it is full screen :+1: 

To answer your question, it could be nice to do the same for the full-screen snippets, so using the option will have a visible effect :slightly_smiling_face: 
But if there are too many, let's not do it, it's not really an issue and the snippets were already like that :+1: 
If you do it, try to do it in one push (without a rebase in it) so I can easily see the diff :slightly_smiling_face: 
And no need to do something proportional, just a really small or no padding at all is fine :+1:

theme_loftspace/views/new_page_template.xml
```diff
@@ -197,12 +197,6 @@
     </xpath>
 </template>
```
sobo-odoo, line 199: Note: Technically, almost all the s_cover of this file should be adapted to remove the padding, since they have the o_three_quarter_height class

But it's not really necessary, they already had a big padding anyway, it's just a nice to have so the option looks good :+1:

sobo-odoo (changes requested):
Some comments :slightly_smiling_face: 
Try to adapt in one push without a rebase in it, so it's easier to see the diff :slightly_smiling_face:

theme_graphene/views/customizations.xml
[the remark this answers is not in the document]
sobo-odoo: Aah yes indeed :+1:

theme_loftspace/views/new_page_template.xml
ptpu-odoo: Added padding to `s_cover` snippet of this theme, so all inheriting new page templates get consistent spacing automatically.

sobo-odoo (approved):
@robodoo r+

PR #1246, opened 2026-03-31
[IMP] theme_bookstore: redesign the theme

ldau-odoo:
| Before | After |
|--------|--------|
| bookstore_screenshot | Home-screenshot-08-03-2026_12_00_PM |

task-6036561

community-PR: https://github.com/odoo/odoo/pull/256839


theme_bookstore/static/description/bookstore_screenshot.webp
anso-odoo, line 1: I think it's better if the screenshot doesn't include the back end arrow

Also, some sections have a white background in the screenshot, while they don't have it in the generated website.

Image

Image

Image

theme_bookstore/views/images.xml
anso-odoo, line 1: I don't think these changes are necessary, they act as a duplicate of what already exists. To check

theme_bookstore/views/global_customizations.xml
```diff
@@ -0,0 +1,366 @@
...
+<odoo>
+
+    <!-- ======== CAROUSEL INTRO ======== -->
+    <template id="s_carousel_intro" inherit_id="website.s_carousel_intro">
+        <!-- Shape -->
+        <xpath expr="//div[hasclass('carousel-inner')]" position="before">
+            <div class="o_we_shape o_html_builder_Bold_17" style="background-image: url(&quot;/html_editor/shape/html_builder%2FBold%2F17.svg?c5=o-color-2&amp;flip=xy&quot;); background-position: 50% 50%;" contenteditable="false"></div>
+        </xpath>
+        <xpath expr="//div[hasclass('s_carousel_intro')]" position="attributes">
+            <attribute name="class" add="s_carousel_boxed s_carousel_rounded bg-white o_cc2" remove="s_carousel_default carousel-dark" separator=" "/>
```
anso-odoo: Not sure why you need "bg-white" here but when we try to achieve a result, we should always do it the way the editor would do it. For example here, instead of using `bg-white o_cc2`, it would be better to use `o_cc1` directly. Also, any `o_ccx` class should be accompanied by an `o_cc` class

theme_bookstore/views/global_customizations.xml
```diff
@@ -0,0 +1,366 @@
...
+            <attribute name="data-oe-shape-data">{"shape":"html_builder/Bold/17","colors":{"c5":"o-color-2"},"flip":["y","x"],"showOnMobile":false,"shapeAnimationSpeed":"0","selectedColor":true}</attribute>
+            <attribute name="style">--position: relative;</attribute>
```
anso-odoo: Not sure this is necessary. Especially as a custom property like this

theme_bookstore/views/global_customizations.xml
```diff
@@ -0,0 +1,366 @@
...
+        <!-- All slides: remove o_cc o_cc1 -->
+        <xpath expr="//div[hasclass('s_carousel_intro_item')]" position="attributes">
+            <attribute name="class" remove="o_cc o_cc1" separator=" "/>
+        </xpath>
+        <xpath expr="//div[hasclass('s_carousel_intro_item')][2]" position="attributes">
+            <attribute name="class" remove="o_cc o_cc1" separator=" "/>
+        </xpath>
+        <xpath expr="//div[hasclass('s_carousel_intro_item')][3]" position="attributes">
+            <attribute name="class" remove="o_cc o_cc1" separator=" "/>
+        </xpath>
```
anso-odoo: If you really want the background of the carousel to be white (cf: `bg-white` comment), i think this can be dropped

theme_bookstore/views/global_customizations.xml
```diff
@@ -0,0 +1,366 @@
...
+        <!-- Slider 1 -->
+        <xpath expr="//div[hasclass('o_grid_item')][1]" position="replace" mode="inner">
```
anso-odoo: Can you check to limit the use of text customisations ? It's not necessary since it will be replaced by AI in the website configurator

theme_bookstore/views/global_customizations.xml
```diff
@@ -0,0 +1,366 @@
...
+    <!-- ======== ATTRIBUTES HORIZONTAL ======== -->
+    <template id="s_attributes_horizontal" inherit_id="website.s_attributes_horizontal">
+        <xpath expr="//section" position="attributes">
+            <attribute name="class" add="o_cc o_cc2" separator=" "/>
+        </xpath>
+    </template>
+
+    <!-- ======== BENTO GRID ======== -->
+    <template id="s_bento_grid" inherit_id="website.s_bento_grid">
+
```
anso-odoo: No need of empty lines

theme_bookstore/views/global_customizations.xml
```diff
@@ -0,0 +1,366 @@
...
+        <!-- Grid container -->
+        <xpath expr="//div[hasclass('o_grid_mode')]" position="attributes">
+            <attribute name="data-row-count">11</attribute>
+            <attribute name="style">gap: 24px 24px;</attribute>
```
anso-odoo: Can't we do `gap: 24px` ?

theme_bookstore/views/global_customizations.xml
```diff
@@ -0,0 +1,366 @@
...
+            <attribute name="style" add="grid-area: 6 / 9 / 12 / 13;" separator=";" />
+        </xpath>
+        <xpath expr="//div[hasclass('o_grid_item')][5]//blockquote" position="replace"/>
+        <xpath expr="//div[hasclass('o_grid_item')][5]//div[hasclass('o_we_bg_filter')]" position="after">
+            <h2 class="h3-fs" style="text-align: center;"><br/>Classics</h2>
+            <p class="lead" style="text-align: center;"><span class="o_small-fs">Discover<strong class="text-o-color-1"> -&gt;</strong></span></p>
+        </xpath>
+    </template>
+
+    <!-- ======== CATEGORY AND PRODUCTS ======== -->
```
anso-odoo, line 154: Since it's a new snippet, maybe you can implement it closer to you needs for Bookstore, so you don't have to do all the custos

theme_bookstore/views/global_customizations.xml
```diff
@@ -0,0 +1,366 @@
...
+    </template>
+
+    <!-- ======== ANNOUNCEMENT SCROLL ======== -->
+    <template id="s_announcement_scroll" inherit_id="website.s_announcement_scroll">
+        <xpath expr="//section" position="attributes">
+            <attribute name="style">--marquee-animation-speed: 5s; --marquee-item-font-size: 40;</attribute>
+        </xpath>
+    </template>
+
+    <!-- ======== PRODUCTS & ADVANTAGES ======== -->
```
anso-odoo, line 183: Same comment as Category and Products

anso-odoo (changes requested):
Hello @ldau-odoo

Nice work,
Some little things to check!

Thanks

PR #1250, opened 2026-04-08
[IMP] theme_anelusia: adapt to new img indicators for gallery

robinlej:
Now, the gallery indicators are `img` tags instead of using background images. This is done to allow the translation of the indicator images.

task-3626918


FrancoisGe:
@robodoo delegate=loco-odoo

loco-odoo:
@robodoo r+

PR #1254, opened 2026-04-14
[ADD] theme_brutalist: add 'Brutalist' new website design-theme

anso-odoo:
This PR introduces the new 'Brutalist' theme to our library of themes.

Community PR: https://github.com/odoo/odoo/pull/259198

task-6069356


theme_brutalist/static/src/js/tour.js
```diff
@@ -0,0 +1,42 @@
...
+
+import * as wTourUtils from '@website/js/tours/tour_utils';
+
+const snippets = [
+    {
+        id: 's_framed_intro',
+        name: 'Framed Intro',
+        groupName: "Intro",
+    },
+    {
```
sobo-odoo, line 16: You forgot the `s_awards_list` snippet :slightly_smiling_face:

theme_brutalist/views/homepage_customizations.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<!-- Framed Intro -->
```
sobo-odoo: If we are back to making big files with many customizations, then I would prefer to have a more visible separator for each snippet, because with the other comments inside, it's hard to see when we are not in the snippet anymore :thinking: 
The old templates were like this 
```xml
<!-- ======== SNIPPET NAME ======== -->
```

It was easy to spot with the "=" and the capital letters, maybe let's use it ? :slightly_smiling_face: (at least capital letters)

theme_brutalist/views/homepage_customizations.xml
```diff
@@ -0,0 +1,42 @@
...
+<template id="configurator_s_framed_intro" inherit_id="website.configurator_s_framed_intro" name="Brutalist s_framed_intro">
```
sobo-odoo: Could be nice to not have the jumbotron text in the first block of the configurator, maybe customize it ? :slightly_smiling_face: 

image

theme_brutalist/views/homepage_customizations.xml
```diff
@@ -0,0 +1,42 @@
...
+    <!-- Heading -->
+    <xpath expr="//h1" position="replace" mode="inner"><span style="font-size: clamp(8px, 1em + 7.6402vw, 400px);">Emma<br/>Sophia Lee</span></xpath>
```
sobo-odoo: Go to the line for the `<span>` tag so it's not on the same line as the xpath :+1:

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,140 @@
...
+    </xpath>
+    <!-- Heading -->
+    <xpath expr="//div[hasclass('o_grid_item')][1]" position="attributes">
+        <attribute name="class" add="g-col-lg-8 g-height-13 col-lg-8" remove="g-col-lg-6 g-height-10 col-lg-6" separator=" "/>
+        <attribute name="style">--grid-item-padding-y: 0; --grid-item-padding-x: 0; grid-area: 1 / 1 / 14 / 9; order: 2; z-index: 1;</attribute>
+    </xpath>
+    <xpath expr="//h1" position="attributes">
+        <attribute name="class" remove="display-4" separator=" "/>
+    </xpath>
+    <xpath expr="//h1" position="replace" mode="inner"><span style="font-size: clamp(8px, 1em + 6vw, 400px);">Emma<br/>Anderson</span></xpath>
```
sobo-odoo: Same here, move the span tag to the line :slightly_smiling_face: (or is there a reason to put it on the same line ?)

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,140 @@
...
+        <attribute name="style">--grid-item-padding-y: 0; --grid-item-padding-x: 0; grid-area: 3 / 9 / 6 / 13; order: 3; z-index: 3;</attribute>
+    </xpath>
+    <xpath expr="//p[hasclass('lead')]" position="attributes">
+        <attribute name="style">text-align: end;</attribute>
+    </xpath>
+    <!-- CTA -->
+    <xpath expr="//p[last()]" position="replace"/>
+    <!-- Image -->
+    <xpath expr="//div[hasclass('o_grid_item_image')]" position="attributes">
+        <attribute name="class" add="g-col-lg-3 g-height-8 col-lg-3 o_we_padding_highlight border" remove="g-col-lg-5 g-height-14 col-lg-5 o_cc o_cc2 order-lg-0" separator=" "/>
```
sobo-odoo: - You shouldn't remove the `order-lg-0` class, since you keep the `order` style :+1: 
- Remove the  `o_we_padding_highlight` class, this is the temporary preview class when using the grid padding option :wink:

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,140 @@
...
+        <attribute name="style">--grid-item-padding-y: 0; --grid-item-padding-x: 0; --box-border-left-width: 16px; --box-border-bottom-width: 16px; --box-border-right-width: 16px; --box-border-top-width: 16px; grid-area: 6 / 10 / 14 / 13; border-left-color: var(--secondary) !important; border-bottom-color: var(--secondary) !important; border-right-color: var(--secondary) !important; border-top-color: var(--secondary) !important; order: 1; z-index: 2;</attribute>
```
sobo-odoo: You can do `border-color: var(--secondary) !important;` directly, the option understands it :wink: 

And in general, same for `border-with`, you only need the variables if there is an image or filter that should follow the rounding (so here keep the variables for the width, because if we use the rounding option of the column, it will be weird) :+1:

theme_brutalist/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Part of Odoo. See LICENSE file for full copyright and licensing details.
```
sobo-odoo, line 1: (Putting the comment at a random place)
Compared to existing themes, we are missing these files/changes:
- the  `LICENSE` file
- adding the theme in the `.weblate.json` file
- exporting the theme `.pot` file and put it in the i18n folder 

For the last two points, I asked the translation team and we have to do it ourselves :+1: 
For the License, we have to put one (I asked JKE), but we could have only one file at the root of design theme instead of in each theme :slightly_smiling_face:

theme_brutalist/__manifest__.py
```diff
@@ -0,0 +1,46 @@
+{
+    'name': 'Brutalist Theme',
+    'description': 'Brutalist Theme - Raw and Bold Bootstrap Theme for Odoo CMS',
+    'category': 'Theme/Creative',
+    'summary': 'Brutalist, Bold, Raw, Monospace, High Contrast, Modern, Minimal',
+    'sequence': 170,
```
sobo-odoo, line 6: Question: can you explain to me what is the sequence ? :slightly_smiling_face:

theme_brutalist/__manifest__.py
```diff
@@ -0,0 +1,46 @@
...
+        'static/description/brutalist_description.webp',
+        'static/description/brutalist_screenshot.webp',
+    ],
+    'configurator_snippets': {
+        'homepage': ['s_framed_intro', 's_awards_list', 's_title', 's_wavy_grid', 's_freegrid', 's_image_punchy'],
+    },
+    'configurator_snippets_addons': {
+        'website_sale': {
+            'homepage': [
+                ('website_sale.s_dynamic_snippet_category_list', 'after', 's_framed_intro'),
```
sobo-odoo: I would not put it as the second snippet, I think it ruins the page a bit :open_mouth: 
After free grid is nice, it doesn't ruin the flow of the page (at least put it after awards list) :slightly_smiling_face:

theme_brutalist/__manifest__.py
```diff
@@ -0,0 +1,46 @@
+{
+    'name': 'Brutalist Theme',
+    'description': 'Brutalist Theme - Raw and Bold Bootstrap Theme for Odoo CMS',
+    'category': 'Theme/Creative',
+    'summary': 'Brutalist, Bold, Raw, Monospace, High Contrast, Modern, Minimal',
```
sobo-odoo, line 5: Question: Not sure how the industries work exactly, but shouldn't there be words here more linked to some of them ? (like anelusia mentioning clothes, or aviato with travel)

I mean, what kind of things can a user write to the configurator to end up with this theme recommended ?

theme_brutalist/models/theme_brutalist.py
```diff
@@ -0,0 +1,14 @@
...
+    _inherit = 'theme.utils'
+
+    def _theme_brutalist_post_copy(self, mod):
+        self.disable_view('website.option_header_brand_logo')
+        self.disable_view('website.header_text_element')
+        self.disable_view('website.header_search_box')
+        self.disable_view('website.header_call_to_action')
+
+        self.enable_view('website.template_footer_headline')
+        self.enable_view('website.footer_no_copyright')
```
sobo-odoo, line 14: We will need the python support for that (already ok for `template_footer_headline`), otherwise it won't be reset correctly from a theme to another :+1: I will do it on the community PR :wink:

theme_brutalist/views/homepage_customizations.xml
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="utf-8"?>
```
sobo-odoo, line 1: When removing  the theme customizations, it was decided to only keep the homepage ones, but we want to see these homepage version in the snippets dialog => so we in fact don't want `configurator_` templates, only general customization that are used both for the theme and the configurator :+1: 

Could you combine this file and the global one, into one global file ? :slightly_smiling_face:

theme_brutalist/static/src/scss/primary_variables.scss
```diff
@@ -0,0 +1,126 @@
...
+
+$brutalist-4: map-get($o-color-palettes, 'brutalist-4');
+$brutalist-4: map-merge($brutalist-4, ('footer': 1));
+$o-color-palettes: map-merge($o-color-palettes, ('brutalist-4': $brutalist-4));
+
+$o-selected-color-palettes-names: append($o-selected-color-palettes-names, 'brutalist-4');
+
+$o-gray-color-palettes: map-merge($o-gray-color-palettes, (
+    'brutalist-4': map-get($o-gray-color-palettes, 'brutalist'),
+));
```
sobo-odoo, line 103: I don't think we need these 3 color palettes, as there are never suggested so we can't use them :+1:

theme_brutalist/static/src/scss/primary_variables.scss
```diff
@@ -0,0 +1,126 @@
...
+        'o-color-5': #000000,
+
+        'o-cc4-btn-primary': 'o-color-2',
+        'o-cc5-btn-primary': 'o-color-2',
+    )
+));
+
+$brutalist: map-get($o-color-palettes, 'brutalist');
+$brutalist: map-merge($brutalist, ('footer': 1));
+$o-color-palettes: map-merge($o-color-palettes, ('brutalist': $brutalist));
```
sobo-odoo: You can just put `footer: 1`  in the fist map-merge directly, no ? Since you are creating the brutalist palette here :slightly_smiling_face: 
So all this become this :arrow_down: 

```scss
$o-color-palettes: map-merge($o-color-palettes, (
    'brutalist': (
        'o-color-1': #000000,
        'o-color-2': #FFFF01,
        'o-color-3': #FDFCFC,
        'o-color-4': #F3F1ED,
        'o-color-5': #000000,

        'o-cc4-btn-primary': 'o-color-2',
        'o-cc5-btn-primary': 'o-color-2',

        'footer': 1,
    )
));
```

Is it normal that this color palette doesn't appear in the suggestions (both in the configurator and in the theme tab) ?

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,140 @@
...
+    <xpath expr="//section" position="attributes">
+        <attribute name="data-oe-shape-data"/>
+    </xpath>
+    <!-- Shape -->
+    <xpath expr="//div[hasclass('o_we_shape')]" position="replace"/>
+    <!-- Grid -->
+    <xpath expr="//div[hasclass('row')]" position="attributes">
+        <attribute name="data-row-count">28</attribute>
+    </xpath>
+    <!-- Heading -->
```
sobo-odoo: Nitpick: instead of heading, call it Column 1 or Grid item 1 ? And then move heading above the xpath touching the h2 

In fact, now that I'm reading the rest, could you keep the structure like 
Grid item 1
Heading 1
Image 1

Grid item 2
Heading 2

It will be easier to review and follow the xpaths :slightly_smiling_face:

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,140 @@
...
+    <xpath expr="//div[hasclass('o_grid_item')][1]" position="attributes">
+        <attribute name="class" add="g-height-5 g-col-lg-6 col-lg-6" remove="g-height-3 g-col-lg-4 col-lg-4" separator=" "/>
+        <attribute name="style">--grid-item-padding-y: 40px; grid-area: 12 / 4 / 17 / 10; z-index: 1;</attribute>
+    </xpath>
+    <xpath expr="//h2" position="attributes">
+        <attribute name="class" add="display-3-fs" separator=" "/>
+    </xpath>
+    <xpath expr="//h2" position="replace" mode="inner"><em>Designed</em> for<br/><font class="bg-o-color-2">Modern Living</font></xpath>
```
sobo-odoo, line 69: Nitpick: you could replace it in one xpath, instead of doing one xpath for the class and one for the content

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,140 @@
...
+    <!-- Image 1 -->
+    <xpath expr="(//img)[1]" position="attributes">
+        <attribute name="class" remove="rounded" separator=" "/>
+    </xpath>
+    <!-- Caption 1 -->
+    <xpath expr="(//div[hasclass('o_grid_item')])[3]" position="attributes">
+        <attribute name="class" add="g-height-1" remove="g-height-3" separator=" "/>
+        <attribute name="style">--grid-item-padding-y: 16px; grid-area: 11 / 1 / 12 / 8; z-index: 3;</attribute>
```
sobo-odoo: Make it a height of 2, technically the text is bigger than one row if we consider the margin :slightly_smiling_face: 
image

I will go through each snippet when the grid mode will be responsive, so catching these cases now already when I see one :wink:

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,140 @@
...
+    <xpath expr="(//div[hasclass('o_grid_item_image')])[2]" position="attributes">
+        <attribute name="style">--grid-item-padding-y: 0; grid-area: 1 / 10 / 10 / 13; z-index: 4;</attribute>
+    </xpath>
+    <xpath expr="(//img)[2]" position="attributes">
+        <attribute name="class" remove="rounded" separator=" "/>
+    </xpath>
+    <!-- Caption 2 -->
+    <xpath expr="(//div[hasclass('o_grid_item')])[5]" position="attributes">
+        <attribute name="class" add="g-height-1" remove="g-height-3" separator=" "/>
+        <attribute name="style">--grid-item-padding-y: 16px; grid-area: 10 / 10 / 11 / 13; z-index: 5;</attribute>
```
sobo-odoo: Same comment about the height of 2 :slightly_smiling_face:

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,140 @@
...
+    <xpath expr="(//div[hasclass('o_grid_item_image')])[3]" position="attributes">
+        <attribute name="style">--grid-item-padding-y: 0; grid-area: 18 / 1 / 27 / 4; z-index: 6;</attribute>
+    </xpath>
+    <xpath expr="(//img)[3]" position="attributes">
+        <attribute name="class" remove="rounded" separator=" "/>
+    </xpath>
+    <!-- Caption 3 -->
+    <xpath expr="(//div[hasclass('o_grid_item')])[7]" position="attributes">
+        <attribute name="class" add="g-height-1" remove="g-height-3" separator=" "/>
+        <attribute name="style">--grid-item-padding-y: 16px; grid-area: 27 / 1 / 28 / 4; z-index: 7;</attribute>
```
sobo-odoo: Same here about height 2

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,140 @@
...
+        <p>Launch Edition</p>
+    </xpath>
+    <!-- Image 4 -->
+    <xpath expr="(//img)[4]" position="attributes">
+        <attribute name="class" remove="rounded" separator=" "/>
+    </xpath>
+    <!-- Caption 4 -->
+    <xpath expr="(//div[hasclass('o_grid_item')])[9]" position="attributes">
+        <attribute name="class" add="g-height-1" remove="g-height-3" separator=" "/>
+        <attribute name="style">--grid-item-padding-y: 16px; grid-area: 28 / 6 / 29 / 13; z-index: 9;</attribute>
```
sobo-odoo: And same here for height 2

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,140 @@
...
+    </xpath>
+    <!-- Section -->
+    <xpath expr="//section" position="attributes">
+        <attribute name="class" add="o_cc1" remove="o_cc2" separator=" "/>
+    </xpath>
+    <xpath expr="//h2" position="replace" mode="inner">
+        <span style="font-size: clamp(8px, 1em + 1.657vw, 400px);" class="o_rfs">
+            <font class="bg-o-color-2">A PUNCHY<br/></font>HEADLINE
+        </span>
+    </xpath>
```
sobo-odoo: You could do the xpath on the font tag directly no, since you only replace its class ?

sobo-odoo (changes requested):
Here is my review :slightly_smiling_face:  Very nice theme visually :wink: 

Not sure why, but this theme is really bugged functionally,
- when I run the configurator, I end up with a homepage with the default images and the default palette, not the theme ones
- the palette is never suggested, not in the configurator and not in the theme tab

chgo-odoo:
> when I run the configurator, I end up with a homepage with the default images and the default palette, not the theme ones

FYI there is a known issue, I think it’s the one you’re mentioning, if you install only website on a DB you should be able to see the theme customized by the configurator, so the theme should not be preinstalled on the DB

theme_brutalist/__manifest__.py
chgo-odoo: From what I understood when creating the boilerplate we used for our themes it's a value controlling the display order of themes when the user reaches the screen "Choose your theme". That being said, the value we used there is totally arbitrary, I don't know if we should pay that much attention to it (?)

theme_brutalist/views/homepage_customizations.xml
chgo-odoo: I never heard about that honestly, are you refering to the PR DUAU is working on ? 😅

AFAIK we were requested to split customizations across two different files because I might define a huge padding for a specific snippet in a `configurator_` template due to the snippets sequence, but I might not want this huge padding in the dialog. Same goes for a `o_cc` background and so on ...

cc @stefanorigano

sobo-odoo: Yes, it's that PR, I discussed this morning again with Lionel, and this the reached conclusion 

I personally agree with you, because I know we have base custo and then really fancy ones for the configurator with more dramatic shapes, padding and title so it would be weird to propose them in the dialog, but the reason is that if a user saw the block and finds it beautiful, he should be able to reuse it easily and find it in the dialog :shrug:

But I guess that if the only think you are modifying is the padding, we could keep two files, to be less dramatic in the dialog :thinking: But for the style and content itself, it should be the same in the dialog and on the page :slightly_smiling_face:

stefanorigano: Confirmed @sobo-odoo , we need a mix of both :)

There are certain changes that **cannot** be implemented in the base template, such as text adaptations. If applied globally, these changes would be incorrectly retained even if they don’t match the industry selected by the user via the configurator (see https://github.com/odoo/design-themes/pull/1260).

Other visual changes are purely contextual for the homepage sequence and would not make sense if applied globally. For example, a shape that matches the one in the following snippet, or the cc_[x] classes applied to the entire homepage because makes sense for that sequence (only).

Decorative element or other style elements that define the "style" of the snippets itself, regardless by the context,  are applied globally instead 👍 

The line between customization and arbitrary changes is a bit blurry, I understand.
Our approach is as follows:
- If a customization defines the style of the snippet, it goes into the “global” category.
- If it makes sense only in the context of the homepage, such as stuff that could have presumably been made by a human using the editor to achieve a cohesive result,  thenit goes into the “homepage”.

sobo-odoo: I think we need to discuss it together once and for all then, because Lionel said that "if I see it and I like it, I need to easily find it" (not his words, just a summary of the idea)
I totally agree with you, but since we are removing all the customizations, it doesn't make sense to have some customized blocks that are not the homepage ones :thinking: 
I agree to separate for the text content and the padding, but I think that the shapes are parts of the things they want to keep since it is really visual :thinking: 

Maybe to discuss with LEBL and BOJE one last time so we are fixed on the direction ? (I personally don't want the customizations to be removed at all, especially now since we removed the new page template nightmare :smiling_face_with_tear:)

stefanorigano: > "...I need to easily find it"

Well, you will find it, except:
- the industry tailored text
- the padding that has been adapted for that context
- "connection" shapes (the decorative ones instead stay global)

theme_brutalist/views/homepage_customizations.xml
chgo-odoo: Seems to be the case already, at least after testing this morning with the new configurator

theme_brutalist/static/src/scss/primary_variables.scss
chgo-odoo: They will be displayed in the theme tab, I think there was an issue that made them not visible but they are indeed displayed there.

**EDIT:** Seems to be related to [what I pointed out here](https://github.com/odoo/design-themes/pull/1254#issuecomment-4709617901), if you start a db with website and goes through the configurator to install the theme you'll see the palette but once the theme is installed it disappears

theme_brutalist/static/src/scss/primary_variables.scss
chgo-odoo: Yes, it's the same issue as the comment above

theme_brutalist/views/global_customizations.xml
chgo-odoo: Yes, I see there is the addition of the `<br/>` as well but it doesn't seem to be needed (?)

chgo-odoo:
Hello @sobo-odoo, thanks for the detailed review (I shared generic feedbacks with the team so we can (hopefully) apply them everywhere.

I added a `!fixup` on top and rebased the branch so we have the new configurator. It should make the review easier

Thanks !

theme_brutalist/static/src/scss/primary_variables.scss
```diff
@@ -0,0 +1,130 @@
...
+
+        'o-cc4-btn-primary': 'o-color-2',
+        'o-cc5-btn-primary': 'o-color-2',
+
+        'footer': 1,
+    )
+));
+
+$brutalist: map-get($o-color-palettes, 'brutalist');
+$o-color-palettes: map-merge($o-color-palettes, ('brutalist': $brutalist));
```
sobo-odoo: You can remove this, you already merged it when creating the palette normally :wink: (same below)

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,141 @@
...
+    <!-- Heading -->
+    <xpath expr="//div[hasclass('o_grid_item')][1]" position="attributes">
+        <attribute name="class" add="g-col-lg-8 g-height-13 col-lg-8" remove="g-col-lg-6 g-height-10 col-lg-6" separator=" "/>
+        <attribute name="style">--grid-item-padding-y: 0; --grid-item-padding-x: 0; grid-area: 1 / 1 / 14 / 9; order: 2; z-index: 1;</attribute>
+    </xpath>
+    <xpath expr="//h1" position="attributes">
+        <attribute name="class" remove="display-4" separator=" "/>
+    </xpath>
+    <xpath expr="//h1" position="replace" mode="inner">
+        <span style="font-size: clamp(8px, 1em + 6vw, 400px);">Emma<br/>Anderson</span>
```
sobo-odoo, line 19: Shouldn't you update the theme screenshot and description, since it is not Emma Sophia Lee anymore ?

theme_brutalist/__manifest__.py
```diff
@@ -0,0 +1,46 @@
+{
+    'name': 'Brutalist Theme',
+    'description': 'Minimal header opens onto a personal-brand hero pairing oversized serif typography with a yellow-framed portrait, then an awards list, a centered statement, and asymmetric black-and-white image grids before a full image wall and closing visual CTA, with yellow highlight accents as the recurring visual signature. Editorial and image-forward, portfolio-driven / suited for architecture firms, photographers, design studios, and personal portfolios',
+    'category': 'Theme/Creative',
+    'summary': 'Brutalist, Bold, Raw, Monospace, High Contrast, Modern, Minimal',
```
sobo-odoo, line 5: I still think you should add some specific words here, like "Photography, Architecture, Design" so industries can suggest it (it is still based on the summary currently) :slightly_smiling_face:

theme_brutalist/static/src/scss/primary_variables.scss
```diff
@@ -0,0 +1,130 @@
...
+
+$o-selected-color-palettes-names: append($o-selected-color-palettes-names, 'brutalist');
+
+// Second palette
```
sobo-odoo, line 47: When you say "we are aware of an issue",  is the issue happening on all themes or only on the new ones ? (to know if I can merge as is or if I have to dig haha) I never saw it before :open_mouth: 

Since when does it happen ? So I can ask someone to fix it if it's not already the case :+1:

theme_brutalist/views/global_customizations.xml
```diff
@@ -0,0 +1,141 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<!-- ======== FRAMED INTRO ======== -->
+<template id="s_framed_intro" inherit_id="website.s_framed_intro" name="Brutalist s_framed_intro">
```
sobo-odoo, line 5: About the jumbotron, I was talking about the screenshots :slightly_smiling_face: 
I think the text is generated in the configurator so we will never see it there, but we can still see it when switching themes 

Surely a nitpick, but since you changed the text in s_title, I think it could be nice to do it for the first block too :slightly_smiling_face:

sobo-odoo (changes requested):
Some last comments :+1:

theme_brutalist/views/global_customizations.xml
chgo-odoo: I've to admit that this is a nitpick to me 😬 

It makes sense for the title because we need a huge amount of text to achieve the desired layout (we could even question the need of this xpath actually, with the new `s_text_bold` snippet)

theme_brutalist/static/src/scss/primary_variables.scss
chgo-odoo: Briefly checked, it seems to be the same issue as the image. Cannot exactly say since when it happens but recently for sure, we didn't notice it before. I think it was spotted one or two weeks ago

You can test it by starting a design-theme DB with only website installed
Create a new website => choose the brutalist theme, then in the editor change the palette => it appears

Now you create a new website (so the module `theme_brutalist` is installed), you choose it again => palette do not appear, and images are the theme default ones.

It's not a side effect of GUJA's PR as it was already there before AFAIK but I don't have much more context.

Also there seems to be an issue with the themes palettes defined in `primary_variables.scss`, e.g the beauty ones, when I run the configurator the first time then go in the color palette changes, I don't even see them. I don't know if that's related or not

sobo-odoo (approved):
It's okay with me, really nice theme :slightly_smiling_face: 

@robodoo r+

PR #1256, opened 2026-04-18
[IMP] themes_*: cleanup `new_page_template`

chgo-odoo:
- requires:
   - https://github.com/odoo/odoo/pull/259990
   - https://github.com/odoo/enterprise/pull/115215
   - https://github.com/odoo/upgrade/pull/10046
   - https://github.com/odoo/documentation/pull/17550

task-6054072


test_themes/tests/test_new_page_templates.py
```diff
@@ -129,31 +129,16 @@
...
-            views = self.env['ir.ui.view'].search([
-                ('key', 'like', f'{website.theme_id.name}.new_page_template%_s_'),
-            ])
-            if website.theme_id.name != 'theme_default':
-                self.assertGreater(len(views), 10, "Test should have encountered some views in theme %r" % website.name)
-            for view in views:
-                self.assertEqual(view.mode, 'extension', "Theme's new page template customization %r should never be primary" % view.key)
-                name = view.key.split('.')[1]
-                parent_name = view.inherit_id.key.split('.')[1]
-                self.assertEqual(name, parent_name, "Theme's new page template customization %r should use the same name as their parent %r" % (view.key, view.inherit_id.key))
```
sobo-odoo, line 144: This test is the general one so it is also testing the default theme and the logic of the new page template, you should therefore not remove it but adapt it a bit instead :slightly_smiling_face:

I would say to just remove these lines:
```py
if website.theme_id.name != 'theme_default':
    self.assertGreater(len(views), 10, "Test should have encountered some views in theme %r" % website.name)
```

(Also to be future-proof, in case we add back some specific templates for each theme in the future, or want some customizations for some pages :smiling_face_with_tear:)

test_themes/tests/test_new_page_templates.py
```diff
@@ -129,31 +129,16 @@
...
         view_ids = set()
         websites_themes = self.env['website'].get_test_themes_websites()
         for website in websites_themes:
             with MockRequest(self.env, website=website):
                 views = self.env['ir.ui.view'].search([
-                    '|', '|',
+                    '|',
                     ('key', 'like', f'{website.theme_id.name}.s_'),
                     ('key', 'like', f'{website.theme_id.name}.configurator'),
-                    ('key', 'like', f'{website.theme_id.name}.new_page'),
```
sobo-odoo: Same here, there are new page templates in default and this is the general test, so you should still consider them (there just won't be any new page template to test in the loop for the themes that's all) :slightly_smiling_face:

test_themes/tests/test_new_page_templates.py
```diff
@@ -129,31 +129,16 @@
 @tagged('post_install', '-at_install')
```
sobo-odoo, line 129: Commit message: some lines exceed 72 chars :+1:

test_themes/tests/test_new_page_templates.py
```diff
@@ -374,7 +359,7 @@ def check(theme_name, website):
         for website in websites_themes:
             view_count += check(website.name, website)
         _logger.info("Tested %s views", view_count)
-        self.assertGreater(view_count, 2900, "Test should have checked many views")
+        self.assertGreater(view_count, 1000, "Test should have checked many views")
```
sobo-odoo: I guess the count was adapted according to how the test was failing, but when I test locally it says `Tested 4959 views` which is way bigger than 2900 (not sure how 2900 was chosen to begin with) so just telling to make sure this number is a bit close to the reality (same for other count updates) :slightly_smiling_face:

theme_nano/views/snippets/s_parallax.xml
```diff
@@ -5,10 +5,7 @@
     <!-- Section -->
     <xpath expr="//section" position="attributes">
         <attribute name="class" add="pt48 pb48" separator=" "/>
-        <attribute name="data-scroll-background-ratio" add="3" remove="1"/>
-    </xpath>
-    <xpath expr="//div[hasclass('o_we_bg_filter')]" position="attributes">
-        <attribute name="class" add="bg-black-25" remove="bg-black-50" separator=" "/>
+        <attribute name="data-scroll-background-ratio" add="3" remove="-1.5"/>
```
sobo-odoo, line 8: Depending on what you do based on my comment on the community PR, don't forget to adapt here :wink:

theme_notes/views/snippets/s_parallax.xml
```diff
@@ -7,7 +7,7 @@
         <attribute name="data-oe-shape-data">{"shape":"html_builder/Rainy/08_001","flip":[]}</attribute>
     </xpath>
     <!-- Shape -->
-    <xpath expr="//div[hasclass('o_we_bg_filter')]" position="after">
+    <xpath expr="//span[hasclass('s_parallax_bg_wrap')]" position="after">
```
sobo-odoo: Same here :wink:

sobo-odoo (changes requested):
Some comments on the test (+ some reminders) :slightly_smiling_face:

test_themes/tests/test_new_page_templates.py
chgo-odoo: I restored them in the end, not sure if you suggested to raise the count ? If yes I'll repush

theme_nano/views/snippets/s_parallax.xml
```diff
@@ -5,10 +5,9 @@
     <!-- Section -->
     <xpath expr="//section" position="attributes">
         <attribute name="class" add="pt48 pb48" separator=" "/>
-        <attribute name="data-scroll-background-ratio" add="3" remove="1"/>
     </xpath>
     <xpath expr="//div[hasclass('o_we_bg_filter')]" position="attributes">
-        <attribute name="class" add="bg-black-25" remove="bg-black-50" separator=" "/>
+        <attribute name="data-scroll-background-ratio" add="3" remove="-1.5"/>
     </xpath>
```
sobo-odoo, line 9: I think you didn't do the xpath correctly: the `data-scroll-background-ratio` xpath should be on the section, and you should completely remove the bg_filter xpath :slightly_smiling_face:

theme_odoo_experts/views/snippets/s_references.xml
```diff
@@ -16,7 +16,9 @@
         <attribute name="data-row-count">2</attribute>
         <attribute name="style" add="column-gap: 24px;"/>
     </xpath>
+</template>
 
+<template id="s_references_columns" inherit_id="website.s_references_columns">
```
sobo-odoo, line 21: I didn't think about the custo when I said to extract the columns :grimacing: Because then you have to customize `s_references_lite` to set it in grid mode too otherwise it looks like this (and is not correct since the columns are grid ones, but the grid mode is not set)

image

So two choices: either you customize references_lite too, or you remove/undo the extraction of the columns :slightly_smiling_face:

theme_zap/views/snippets/s_references.xml
```diff
@@ -16,6 +16,9 @@
     </xpath>
     <!-- Link -->
     <xpath expr="//a" position="replace"/>
+</template>
+
+<template id="s_references_columns" inherit_id="website.s_references_columns">
```
sobo-odoo, line 21: Same comment here, the references lite of this theme should be customized too (if you don't undo the extraction)

image

sobo-odoo (changes requested):
Some details :+1:

theme_nano/views/snippets/s_parallax.xml
chgo-odoo: Woops 😬

theme_odoo_experts/views/snippets/s_references.xml
```diff
@@ -16,7 +16,9 @@
         <attribute name="data-row-count">2</attribute>
         <attribute name="style" add="column-gap: 24px;"/>
     </xpath>
+</template>
```
sobo-odoo, line 19: There is the issue outside of your PR too but I just noticed it: it seems that this customization doesn't apply on the references snippet :open_mouth: 

If you check on runbot, it is not in grid at all (like it is completely ignored) and in your PR, the part here is ignored, but the columns below are well transformed to grid ones

Could you check why ? :slightly_smiling_face:

sobo-odoo (changes requested):
Small detail here :slightly_smiling_face: 
The issue was here before, but now that the columns are well customized, we have to fix it :+1:

theme_odoo_experts/views/snippets/s_references.xml
chgo-odoo: I noticed this as well on runbot indeed... But local seems to work fine 🤔 (I just installed test_themes)
image I don't know what could be the difference between both setups, an extra module seems weird but I'll rebase the branch anyway after fixing the comments, we'll see

chgo-odoo: Found the issue actually... Both `s_references` and `s_references_grid` were sharing the same id... Not sure exactly why it's working fine locally but I guess on Runbot `s_references_grid` is overriding the xpath from `s_references`, like a cascade while locally each file is processed independently

chgo-odoo:
@sobo-odoo I added two commits on top to fix issues unrelated to new page templates but I spotted them while checking the console in the Odoo Experts theme

theme_odoo_experts/__manifest__.py
```diff
@@ -62,7 +62,7 @@
         'static/description/odoo_experts_screenshot.webp',
```
sobo-odoo, line 61: Commit message:
- make the title start with a verb, like "fix various issues" or something like that :wink: 
- maybe also mention the issue about the `s_mockup_image` you are fixing :slightly_smiling_face:

theme_artists/views/snippets/s_numbers_boxed.xml
```diff
@@ -4,7 +4,7 @@
 <template id="s_numbers_boxed" inherit_id="website.s_numbers_boxed">
```
sobo-odoo, line 4: Commit message:
- nitpick: `but not visible within the editor` -> I would say `but not considered as applied by the option` instead, it is more clear on what the issue is, at first I thought it meant it was not visible in edit mode haha

theme_bistro/views/snippets/s_wavy_grid.xml
```diff
@@ -5,10 +5,6 @@
     <!-- Layout -->
     <xpath expr="//section" position="attributes">
         <attribute name="class" add="o_cc o_cc5" separator=" "/>
-        <attribute name="data-oe-shape-data">{'shape':'html_builder/Bold/26','colors':{'c5': 'o-color-2'},'flip':[],'showOnMobile':false,'shapeAnimationSpeed':'0'}</attribute>
-    </xpath>
-    <xpath expr="//div[hasclass('o_we_shape')]" position="replace">
-        <div class="o_we_shape o_html_builder_Bold_26" style="background-image: url('/html_editor/shape/html_builder/Bold/26.svg?c5=o-color-2');"/>
```
sobo-odoo, line 11: Maybe add another shape instead => with the default shape, the text is not visible when the shape is behind it
image

Or maybe just change the shape color in the theme ? `o-color-1` looks nice :slightly_smiling_face: 
image

theme_odoo_experts/views/snippets/s_image_title.xml
```diff
@@ -5,10 +5,6 @@
     <xpath expr="//section" position="attributes">
         <attribute name="class" add="o_cc1" remove="o_cc5 oe_img_bg" separator=" "/>
         <attribute name="style"/>
-        <attribute name="data-oe-shape-data">{'shape':'html_builder/Bold/26','colors':{'c5':'o-color-3'},'flip':[],'showOnMobile':false,'shapeAnimationSpeed':'0'}</attribute>
-    </xpath>
-    <xpath expr="//div[hasclass('container')]" position="before">
-        <div class="o_we_shape o_html_builder_Bold_26" style="background-image: url('/html_editor/shape/html_builder/Bold/26.svg?c5=o-color-3');"></div>
```
sobo-odoo, line 11: Maybe add another shape for this one, because the block is "naked" like this, it is called "image title" but has no image or color :open_mouth: 
image

In 18.0, the block looked like this 
image

I would say to either do something similar here or to not remove the background image (it depends on the theme image I guess) :slightly_smiling_face:

theme_real_estate/views/snippets/s_wavy_grid.xml
```diff
@@ -5,10 +5,6 @@
     <!-- Layout -->
     <xpath expr="//section" position="attributes">
         <attribute name="class" add="o_cc o_cc2" separator=" "/>
-        <attribute name="data-oe-shape-data">{'shape':'html_builder/Bold/26','colors':{'c5':'o-color-4'},'flip':[],'showOnMobile':false,'shapeAnimationSpeed':'0'}</attribute>
-    </xpath>
-    <xpath expr="//div[hasclass('o_we_shape')]" position="replace">
-        <div class="o_we_shape o_html_builder_Bold_26" style="background-image: url('/html_editor/shape/html_builder/Bold/26.svg?c5=o-color-4');"/>
```
sobo-odoo, line 11: We have the opposite issue of bistro haha, now the default shape is not visible because it has the same color as the background

`o-color-2` or `o-color-4` could be nice :slightly_smiling_face: 
| current | o-color-2 | o-color-4 |
|--------|--------|--------|
| image | image | image |

theme_odoo_experts/views/images.xml
```diff
@@ -375,12 +375,6 @@ Check in theme_loftspace's primary_variables.scss and theme.scss -->
     <field name="name">website.s_showcase_default_image</field>
     <field name="url">/theme_odoo_experts/static/src/img/snippets/s_showcase_default_image.webp</field>
 </record>
-<record id="s_mockup_image_default_image" model="theme.ir.attachment">
-    <field name="key">website.s_mockup_image_default_image</field>
-    <field name="name">website.s_mockup_image_default_image</field>
-    <field name="url">/theme_odoo_experts/static/src/img/snippets/s_mockup_image_default_image.jpg</field>
-</record>
-
```
sobo-odoo, line 383: I think you have to remove the record in an upgrade script too :slightly_smiling_face: (otherwise, the snippets using this image will still use this record, and won't have the image, like your new s_features_card snippet)

But doing `util.remove_record(cr, "theme_odoo_experts.s_mockup_image_default_image")` doesn't seem to work, it doesn't seem to be the right way to remove these images :thinking: 

Do you know how we are supposed to do this ? I can't find existing examples except the code added by this [other PR](https://github.com/odoo/upgrade/pull/10275), but if I'm right, it shouldn't work either :confused:

sobo-odoo (changes requested):
Some comments on your latest changes :slightly_smiling_face: 
Thanks for adding these fixes :wink:

theme_odoo_experts/views/images.xml
chgo-odoo: I genuinely have no idea 😬 

Are we sure this do nothing? Would be weird but I'm far from being an upgrade expert so...

sobo-odoo: I ran the upgrade manually and the record was still there, that's why I think it doesn't work :thinking: 
I will ask people on Monday :slightly_smiling_face:

theme_odoo_experts/views/snippets/s_references.xml
chgo-odoo: @sobo-odoo I guess this should be added in the upgrade script as well as we're renaming the template, right ? If yes, I'm not really sure how as both templates previously shared the same `id`, I'm afraid there is a tiny chance it tries renaming the wrong one

sobo-odoo: When I look at the views in the db (before your PR), there is only one view for s_references, so I think the second one overrides the first one anyway :open_mouth: 
And from my understanding of the upgrades, we remove old things and then the new things are installed automatically so we could remove it without worries (but in practice, the s_mockup  image issues proves me that it is not the case :smiling_face_with_tear:)

Anyway, here logically, we didn't remove anything, so I guess the views will update correctly without doing anything, I will check :wink:

sobo-odoo: I confirm that we don't need to do anything, the views are updated correctly :wink:

theme_odoo_experts/views/images.xml
sobo-odoo: In fact, we don't need an upgrade script at all, sorry about that :open_mouth: I don't know how I tested before, but I checked again and upgrading without script correctly removed the attachment, I surely had the issue because I did your changes manually when testing instead of checking out your PR and I surely did something wrong :grimacing:

sobo-odoo (approved):
@robodoo rebase-ff r+

sobo-odoo:
@robodoo r+

PR #1258, opened 2026-04-21
[FIX] theme_graphene, *: fix background customizations

bvr-odoo:
*: theme_nano

Steps to reproduce:

- Install the "Graphene" theme through the configurator.
- Once on the homepage, you can see that the "Cover" snippet does not have the "oe_img_bg" class on its section. Because of this, its background image repeats.

This commit fixes this case, and a similar one in the "Nano" theme, by adding the "oe_img_bg" class where those snippets are modified with XPath.


FrancoisGe:
@robodoo r+

bvr-odoo:
wrong branch name

PR #1260, opened 2026-04-21
[IMP] scope copy-content to homepages only

stefanorigano:
Currently, themes customize snippets copy globally and in any context.
This is because the configurator was assumed to have, in the future, parsed snippets as well.
However, since this is not the case, users may receive theme-related content despite having selected a completely different business type.
At the current state, homepage content only is parsed by the configurator.

This PR:
- If a snippet is used in the homepage, we move text xpaths there in order to be further customized 
- If the snippet is NOT used in the homepage, we remove the x-path entirely
- We remove xml templates emptied form any xpath
- We bumps the theme's version for migration scripts

Guidelines:
- https://www.odoo.com/odoo/knowledge/59839

Requires:

- https://github.com/odoo/upgrade/pull/10011

task-6152015


stefanorigano:
Will probably handled by @odoo/rd-website , closing this

PR #1268, opened 2026-04-28
[FIX] theme_*: add missing color attributes to shapes

alap-odoo:
[*]=graphene, kea, kiddo, nano, odoo_experts, orchid, treehouse

Before this PR, several shapes were missing color attributes, which prevented the color picker from appearing in the editor.

This PR adds the required color attributes so the color picker is properly available for shapes in the editor. Moreover, these colors are appended in the shape URL as params to ensure that the default colors are correctly applied to the shapes.

task-5880905

### DIFF
--- a/theme_treehouse/views/snippets/s_picture.xml
+++ b/theme_treehouse/views/snippets/s_picture.xml
@@ -8,10 +8,10 @@
     </xpath>
     <!-- Image + Shape -->
     <xpath expr="//img" position="attributes">
-        <attribute name="src">/web_editor/image_shape/website.s_picture_default_image/web_editor/special/special_filter.svg</attribute>
+        <attribute name="src">/web_editor/image_shape/website.s_picture_default_image/web_editor/special/special_filter.svg?c2=o-color-3</attribute>
         <attribute name="data-file-name">01.svg</attribute>
         <attribute name="data-shape">web_editor/special/special_filter</attribute>
-        <attribute name="data-shape-colors">;#046380;;;</attribute>
+        <attribute name="data-shape-colors">;o-color-3;;;</attribute>
     </xpath>
 </template>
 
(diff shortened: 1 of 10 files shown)


robinlej:
@robodoo delegate=bvr-odoo

theme_treehouse/views/snippets/s_picture.xml
```diff
@@ -8,10 +8,10 @@
     </xpath>
     <!-- Image + Shape -->
     <xpath expr="//img" position="attributes">
-        <attribute name="src">/web_editor/image_shape/website.s_picture_default_image/web_editor/special/special_filter.svg</attribute>
+        <attribute name="src">/web_editor/image_shape/website.s_picture_default_image/web_editor/special/special_filter.svg?c2=o-color-3</attribute>
         <attribute name="data-file-name">01.svg</attribute>
         <attribute name="data-shape">web_editor/special/special_filter</attribute>
-        <attribute name="data-shape-colors">;#046380;;;</attribute>
+        <attribute name="data-shape-colors">;o-color-3;;;</attribute>
```
bvr-odoo, line 14: Why are the shape colors defined in hexadecimal in the community PR, while they are defined with variables in the themes? Why this difference?

bvr-odoo (changes requested):

theme_treehouse/views/snippets/s_picture.xml
alap-odoo: We can use hexadecimal values instead of variables, but since variables (especially for image shape) are used in many places within the design themes, I followed the same approach to stay consistent with the codebase.

bvr-odoo (changes requested):
The branch name doesn't match the community branch, so they aren't properly linked.

alap-odoo:
Hello @bvr-odoo 
It's done. Can you please check [here](https://github.com/odoo/design-themes/pull/1274) ?

PR #1269, opened 2026-04-28
Master test adapt custo text guja

guja-odoo:


guja-odoo:
@odoo/rd-runbot Hello :wave: would it be possible to increase the modified files limit for this PR ? thanks !

PR #1279, opened 2026-05-16
[REF] test_themes: fix import for escape_psql

kmagusiak:
Import directly from the module.


sobo-odoo:
@robodoo delegate+

kmagusiak:
@robodoo r+

PR #1288, opened 2026-06-01
[FIX] *: remove deadfiles and flag those dynamically loaded

Julien00859:
The `ManifestLinter._test_data_files` found a bunch of mismatch between the files that are listed in each module's manifest, and those that actually exist on disk.

In this work we fix all the linter's findings.

* All `ir.model.access.csv` files are now loaded in the manifest.
* Files that are dynamically loaded in the code (e.g. `odoo.tools.convert.convert_file`) have been listed in the new `other_files` entry of the manifest.
* Files that once existed in the manifest and on disk, but that have been removed from the manifest in a previous (cleanup) commit have been removed from disk as well.

There are other files that do not fit in those three categories. We reached for the original author/team of those files and decided together what to do.

### DIFF
--- a/theme_nano/i18n/ar.po
+++ b/theme_nano/i18n/ar.po
@@ -7,7 +7,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: Odoo Server saas~18.2\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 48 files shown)


robinlej:
@robodoo delegate+

Julien00859:
@robodoo r+

Julien00859:
Superseded by #1291

PR #1290, opened 2026-06-09
[REF] test_themes: replace useState with proxy

PiaCOS:
In Owl3, uses of `useState` or replace with `proxy`. This commit removes `useState` from the compatibility layer, provides tooling for the migration in `owl3-migration` and removes the last uses of `useState`.


ged-odoo (approved):

robinlej:
@robodoo r+

PR #1294, opened 2026-06-18
[IMP] theme_anelusia: update thumbnail post redesign

chgo-odoo:
Following the redesign of `theme_anelusia` made in Commit[^1], the screenshot of the theme was updated but not its thumbnail. This PR updates the thumbnail to ensure it matches the new theme design.

[^1]: https://github.com/odoo/design-themes/commit/4f68c63261708bf71c49c56ffffdb8d2fa9d5023

task-6313033


chgo-odoo:
@sobo-odoo 👋 

(Can't ping you as reviewer 😶‍🌫️ )

theme_anelusia/static/description/anelusia_description.webp
sobo-odoo, line 1: Oh I just noticed that this ( :arrow_down: )  is not the image anymore, since the image PR :open_mouth: 
image

Could have been nice to keep the images we added in the redesign, they were really nice :open_mouth: (the theme was really cool with them)

The theme screenshot doesn't match the real images then, so I suggest the following:
- either redo the screenshot and say goodbye to the beautiful anelusia theme we redesigned some months ago
- or replace some images used by the configurator snippets by the old ones, to have the same result as when you redesigned it 

What do you think ? :slightly_smiling_face: I prefer the second solution, because Anelusia was really nice :open_mouth: 

And we in fact have to redo all the themes screenshots too then :grimacing:

sobo-odoo (commented):
I just noticed that in fact, no theme match their screenshot anymore, since we replaced all images without keeping some similar to the ones used before :grimacing:

theme_anelusia/static/description/anelusia_description.webp
chgo-odoo: I'm all in for option 2 🥷

sobo-odoo (approved):
Let's go with this one in the end, since we can't restore the old anelusia cover image :slightly_smiling_face: 

@robodoo r+

PR #1296, opened 2026-06-18
[IMP] theme_brutalist: add a missing xpath

chgo-odoo:
In Commit [^1], theme_brutalist` was added to the list of existing themes. Its design makes use of the responsive font-size feature by displaying a heading with minimal content (first name + last name). Since no xpath was defined in the configurator for this content, the configurator could replace the first name + last name with a paragraph or other longer text, resulting in overflow or broken layouts.

To prevent this issue, we added an xpath so that the configurator can determine the appropriate content to insert in that location.

[^1]: https://github.com/odoo/design-themes/commit/0eae6584c089b97a92d36d61f8d02a51e15d125c

task-6315710


sobo-odoo (approved):
@robodoo r+

chgo-odoo:
robodoo retry

PR #1297, opened 2026-06-20
[FIX] theme_muse: fix theme preview

chgo-odoo:
This PR updates the preview for `theme_muse` according to the new `s_about_bold` snippet design. Prior to this commit, the structure of the `s_about_bold` snippet was reworked to avoid using text columns, because they were creating a horizontal scroll, which are visible in the preview as well.

Now that the snippet uses a standard row and col structure, we need to update the preview to match that layout.

task-6318818

### DIFF
--- a/theme_muse/static/description/preview.html
+++ b/theme_muse/static/description/preview.html
@@ -1460,21 +1460,18 @@
 </div>
 </section><section class="s_about_bold pt112 pb88" data-snippet="s_about_bold">
 <div class="container s_allow_columns">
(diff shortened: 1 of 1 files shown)


chgo-odoo:
@sobo-odoo FYI, I forgot to update the preview after we changed the structure of the s_about_bold snippet 😬

sobo-odoo (approved):
Thanks for the fix :slightly_smiling_face: 

@robodoo r+

PR #1298, opened 2026-06-22
[FIX] theme_*: adapt images for new record and missed occurences

mano-odoo:
In theme_graphene we used `landscape_xl_1` but since the shape is
cropping the image to an ultrawide format, ultrawide is more adapted for
this snippet.

theme_enark, theme_monglia, theme_orchid

In https://github.com/odoo/design-themes/commit/d851c17c03f1 we rebinded the themes to use the new records but some occurences of direct file xpath were missed.

task-6321389

### DIFF
--- a/theme_enark/views/snippets/s_cta_box.xml
+++ b/theme_enark/views/snippets/s_cta_box.xml
@@ -21,7 +21,6 @@
     </xpath>
     <!-- Image -->
     <xpath expr="//img" position="attributes">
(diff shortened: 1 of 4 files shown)


mano-odoo:
@odoo/rd-website

FrancoisGe:
@robodoo r+ rebase-ff

sobo-odoo (commented):
@mano-odoo Since you changed many images, don't forget to regenerate the modified themes screenshot and preview so they match :wink: 

(for example, the Orchid hardcoded image is still in the preview)

PR #1300, opened 2026-06-23
[FIX] theme_*: update themes previews

mano-odoo:
Following commit[1] and commit[2], themes were adapted for the new records but the preview weren't updated.

Follow up of
task-6321389

[1]: odoo/design-themes@62f5e1ebd209
[2]: odoo/design-themes@48a964594c91

### DIFF
--- a/theme_enark/static/description/preview.html
+++ b/theme_enark/static/description/preview.html
@@ -1269,7 +1269,7 @@
 <div class="container">
 <div class="s_card o_card_img_horizontal flex-lg-row-reverse card o_cc o_cc5" data-name="Card" data-snippet="s_card" data-vxml="001" style="--card-spacer-x: 64px;--card-spacer-y: 64px;border-width: 0px !important;border-radius: 0px !important">
 <figure class="o_card_img_wrapper ratio ratio-1x1 mb-0">
-<img alt="" class="o_card_img" loading="lazy" src="/theme_enark/static/src/img/snippets/library_image_02.webp"/>
+<img alt="" class="o_card_img" loading="lazy" src="/web/image/website.set_2_square_md_3"/>
 </figure>
 <div class="card-body">
 <h2 class="card-title">
--- a/theme_graphene/static/description/preview.html
+++ b/theme_graphene/static/description/preview.html
@@ -1108,7 +1108,7 @@
 </nav>
 </header>
 <main class="">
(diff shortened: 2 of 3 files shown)


mano-odoo:
@sobo-odoo here is the PR to adapt the previews. @chgo-odoo is handling the screenshots update 😄

theme_enark/static/description/preview.html
```diff
@@ -1219,20 +1226,20 @@ <h2 class="h3-fs">Key Metrics of Company's Achievements</h2>
...
-<img alt="" class="img img-fluid d-block" data-index="1" data-name="Image" loading="lazy" src="/web/image/website.landscape_md_7"/>
-<img alt="" class="img img-fluid d-block" data-index="4" data-name="Image" loading="lazy" src="/web/image/website.set_2_square_md_4"/>
+<img alt="" class="img img-fluid d-block rounded" data-index="1" data-name="Image" loading="lazy" src="/web/image/website.landscape_md_7"/>
+<img alt="" class="img img-fluid d-block rounded" data-index="4" data-name="Image" loading="lazy" src="/web/image/website.set_2_square_md_4"/>
 </div>
 <div class="o_masonry_col o_snippet_not_selectable col-lg-4">
-<img alt="" class="img img-fluid d-block" data-index="2" data-name="Image" loading="lazy" src="/web/image/website.set_2_square_md_6"/>
-<img alt="" class="img img-fluid d-block" data-index="5" data-name="Image" loading="lazy" src="/web/image/website.set_2_square_md_2"/>
+<img alt="" class="img img-fluid d-block rounded" data-index="2" data-name="Image" loading="lazy" src="/web/image/website.set_2_square_md_6"/>
+<img alt="" class="img img-fluid d-block rounded" data-index="5" data-name="Image" loading="lazy" src="/web/image/website.set_2_square_md_2"/>
```
sobo-odoo: Sorry about that but we wrongly removed some images wall customizations when removing the themes customizations :grimacing: I restored them in the PR of CHGO about the theme screenshots :+1: 
So this change here was due to that :+1: 
Could you rebase and redo the previews ? :slightly_smiling_face: 

Note that to keep a consistent CSS between themes, maybe just change the concerned image urls manually instead of regenerating the whole file :slightly_smiling_face: => I think it's better to use the tooling once for all themes at the same time :+1:  we couldn't avoid it for the new themes, but the old ones all have the same base, so I prefer changing only what needs to be (it will be easier to spot issues in previews if they are are consistent) :slightly_smiling_face: 

Also note that the monglia case was fixed in CHGO's [PR ](https://github.com/odoo/design-themes/pull/1302/files#diff-361cbcb69d024d10f90a30e6861130128b2b2c92c54abde1716a855dfa0113f6R1121) :slightly_smiling_face:

sobo-odoo (changes requested):
Small comment :slightly_smiling_face:

mano-odoo:
@sobo-odoo I pushed just the images diff and removed monglia 😇

sobo-odoo (approved):
Thanks :slightly_smiling_face: 

@robodoo r+

PR #1309, opened 2026-07-07
[FIX] theme_brutalist: fix intro snippet font-size in the preview

anso-odoo:
Inside the preview of the theme, the configuration was replacing
entirely the span on which the responsive font-size was set.

We move the responsive font size rule on the h1 itself, which will
never be replaced by the AI

task-6368821

| Before | After |
|--------|--------|
| Screenshot 2026-07-07 at 10 27 24 |  Screenshot 2026-07-07 at 10 30 13 |

### DIFF
--- a/theme_brutalist/static/description/preview.html
+++ b/theme_brutalist/static/description/preview.html
@@ -926,8 +926,7 @@
 <div class="container">
 <div class="row o_grid_mode" data-row-count="13">
 <div class="o_grid_item g-col-lg-8 g-height-13 col-lg-8" style="--grid-item-padding-y: 0; --grid-item-padding-x: 0; grid-area: 1 / 1 / 14 / 9; order: 2; z-index: 1;">
-<h1>
-<span style="font-size: clamp(8px, 1em + 7.6402vw, 400px);">Emma<br/>Sophia Lee</span></h1>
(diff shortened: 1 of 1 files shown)


anso-odoo:
@odoo/rd-website
Hello guys, can you check this PR?
The fix is a bit of a "hack", but it works well and since it's inside `preview.html`, I think it's pretty ok to do it this way. If you see a better way of doing it, I'm all ears

Thanks!

FrancoisGe:
> @odoo/rd-website Hello guys, can you check this PR? The fix is a bit of a "hack", but it works well and since it's inside `preview.html`, I think it's pretty ok to do it this way. If you see a better way of doing it, I'm all ears
> 
> Thanks!

I'm not a fan of editing the `preview.html` files manually because we wanted to automate their refresh in the future. Since it's a no-brainer, I edit a snippet, run the script, and all the HTML is correct based on the current blocks.
@guja-odoo Maybe we can implement the HTML generation script ?

guja-odoo:
Hello !
This will be fixed once this PR https://github.com/odoo/odoo/pull/272715 is merged, so h1 titles can have spans inside without problems :ok_hand:

anso-odoo:
Closing it then, thanks a lot!

PR #1311, opened 2026-07-10
[IMP] theme_graphene: redesign

chgo-odoo:
COM PR: https://github.com/odoo/odoo/pull/275580
task-6377094

| Before | After |
|--------|--------|
| graphene_screenshot | graphene_screenshot |


anso-odoo:
@odoo/rd-website  Hello again, there's also this one ready 🙏

FrancoisGe:
@sobo-odoo or @romo-odoo  Could you check it ? :)

theme_graphene/__manifest__.py
```diff
@@ -19,12 +21,12 @@
         'website.s_cover_default_image': '/theme_graphene/static/src/img/pictures/bg_image_08.webp',
```
sobo-odoo: You can remove `images_preview_theme` :slightly_smiling_face:

theme_graphene/__manifest__.py
```diff
@@ -9,7 +9,9 @@
         'data/generate_primary_template.xml',
```
sobo-odoo, line 9: You didn't update the description :+1:

theme_graphene/static/src/scss/bootstrap_overridden.scss
```diff
@@ -1,3 +0,0 @@
-// Headings font weight
```
sobo-odoo, line 1: Don't delete the file, just empty it :+1:

theme_graphene/data/ir_asset.xml
```diff
@@ -13,10 +13,4 @@
     <path>theme_graphene/static/src/scss/secondary_variables.scss</path>
 </asset>
 
-<asset id="theme_graphene.bootstrap_overridden_scss" name="Bootstrap overridden SCSS">
-    <field name="key">theme_graphene.bootstrap_overridden_scss</field>
-    <bundle directive="prepend">web._assets_frontend_helpers</bundle>
-    <path>theme_graphene/static/src/scss/bootstrap_overridden.scss</path>
-</asset>
-
```
sobo-odoo, line 21: And so don't remove the asset :slightly_smiling_face: (as said in another theme, we will see later how we can clean them safely :+1: )

theme_graphene/models/theme_graphene.py
```diff
@@ -5,9 +5,6 @@ class ThemeUtils(models.AbstractModel):
     _inherit = 'theme.utils'
 
     def _theme_graphene_post_copy(self, mod):
-        self.enable_view('website.template_header_stretch')
+        self.enable_view('website.template_header_default')
```
sobo-odoo: It's not needed to enable the default, it will already be automatically :slightly_smiling_face:

theme_graphene/__manifest__.py
```diff
@@ -19,12 +21,12 @@
...
     },
     'configurator_snippets': {
-        'homepage': ['s_cover', 's_text_image', 's_numbers_grid', 's_mockup_image', 's_comparisons', 's_references'],
+        'homepage': ['s_hero_minimalist', 's_text_block_split', 's_services_grid', 's_showcase_square', 's_references_tiles', 's_parallax', 's_faq_collapse_centered'],
     },
     'configurator_snippets_addons': {
         'website_sale': {
             'homepage': [
-                ('website_sale.s_dynamic_snippet_category_list', 'after', 's_mockup_image'),
+                ('website_sale.s_dynamic_snippet_category_list', 'after', 's_showcase_square'),
```
sobo-odoo, line 26: For the dynamic snippet:
- Given the page design, its background color should be `o_cc1` :slightly_smiling_face: 
- Already there before, but the `d-flex` class should also be removed from `s_dynamic_snippet_title`, adding `d-none` is not enough, so the option is therefore not correct :+1: 
  image

theme_graphene/static/src/scss/primary_variables.scss
```diff
@@ -138,12 +131,3 @@ $o-color-palettes-compatibility-indexes: (
...
 );
-
-//------------------------------------------------------------------------------
-// Shapes
-//------------------------------------------------------------------------------
-
-$o-bg-shapes: change-shape-colors-mapping('html_builder', 'Origins/02_001', (4: 2, 5: rgba(0, 0, 0, 0)));
-$o-bg-shapes: add-extra-shape-colors-mapping('html_builder', 'Origins/02_001', 'second', (4: 2, 5: rgba(0, 0, 0, 0)));
-$o-bg-shapes: change-shape-colors-mapping('html_builder', 'Origins/06_001', (3: 2));
-$o-bg-shapes: change-shape-colors-mapping('html_builder', 'Zigs/01_001', (2: 4));
```
sobo-odoo, line 149: Don't remove this yet, we will clean all the themes at the same time once we are sure it's fine :slightly_smiling_face:

theme_graphene/static/src/scss/primary_variables.scss
```diff
@@ -80,41 +98,16 @@ $o-theme-font-configs: (
...
+    'Geist': (
+        'family': ('Geist', sans-serif),
+        'url': 'Geist:300,300i,400,400i,700,700i',
     ),
 );
 
-$base-1: map-get($o-color-palettes, 'base-1');
-$base-1: map-merge($base-1, ('footer': 5));
-$o-color-palettes: map-merge($o-color-palettes, ('base-1': $base-1));
-
```
sobo-odoo, line 117: Maybe don't remove this, an existing user using this palette with that theme installed will suddenly have his footer color changed :+1: 

In fact, I don't know at all to which extent we have to think to the users... there is a lot of compatibility in this file so I guess we have to be careful, but on the other hand, we can't keep each version of a palette either :grimacing: (same for the fonts)

Do you know what we really are allowed to change or not from a version to another ? :slightly_smiling_face:

theme_graphene/views/global_customizations.xml
```diff
@@ -0,0 +1,48 @@
...
+</template>
+
+<!-- ======== PARALLAX ======== -->
+<template id="configurator_s_parallax" inherit_id="website.configurator_s_parallax">
+    <!-- Section -->
+    <xpath expr="section" position="attributes">
+        <attribute name="data-scroll-background-ratio">3</attribute>
+        <attribute name="class" remove="s_parallax_is_fixed" separator=" "/>
+    </xpath>
+</template>
```
sobo-odoo: It should go in the other file :wink:

theme_graphene/views/global_customizations.xml
```diff
@@ -0,0 +1,48 @@
...
+
+<!-- ======== FOOTER LINKS ======== -->
+<template id="template_footer_links" inherit_id="website.template_footer_links" name="Graphene Footer Links">
+    <!-- Section -->
+    <xpath expr="//section[@data-snippet='s_text_block']" position="attributes">
+        <attribute name="data-oe-shape-data"/>
+    </xpath>
+    <!-- Shape -->
+    <xpath expr="//section[@data-snippet='s_text_block']/div[hasclass('o_we_shape')]" position="replace"/>
+</template>
```
sobo-odoo: You can't remove the footer shape like this, because we won't be able to add a shape ever again on that template anymore (because the xpath will stay active) :+1: 

We should add a way to allow doing this in the future, like enabling a template that will remove the shape, and then the option will also disable that template when we will add a shape :+1: 
We have the same issue about making a footer full-width or small for example :+1: 

We need to think of a system to allow it easily, but it won't be ready for now (for the width, we were supposed to modify the whole footer system to work with templates like the header, but now that I see the shape case, I wonder if we should not do it in another way :thinking:)

theme_graphene/views/global_customizations.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+
+<!-- ======== SERVICES GRID ======== -->
+<template id="s_services_grid" inherit_id="website.s_services_grid" name="Graphene Services Grid">
```
sobo-odoo: Maybe keep the same way of writing the `name` attribute as in other themes to be consistent, so "Theme s_snippet" with the `s_` snippet id => `Graphene s_services_grid` (it will be easier to find the snippets in the Views menu) :slightly_smiling_face: 

Same for the other custos and in the other file :slightly_smiling_face:

theme_graphene/static/description/graphene_screenshot.webp
sobo-odoo, line 1: Just a comment: the hero minimalist image is the same as the parallax one => since they are big highlighted images, maybe change one of them (in one of the blocks customization) so they are not repeated ? :slightly_smiling_face:

sobo-odoo (changes requested):
Here is my review :slightly_smiling_face:

theme_graphene/views/global_customizations.xml
anso-odoo: Keeping it for now then 👍

anso-odoo:
@sobo-odoo Should be mostly fine now, thank!

sobo-odoo (approved):
Thanks for the work :slightly_smiling_face: 

@robodoo r+

PR #1327, opened 2026-07-30
[FIX] tooling: highlight only the active preview menu

bvr-odoo:
Theme previews created all sample menu links with the home URL, so they were all displayed as active. Give each link its own URL so only Home is highlighted on the homepage.

task-6368411


FrancoisGe:
@robodoo r+

PR #1333, opened 2026-08-20
[FIX] theme_bistro: regenerate preview with stable Material icons URLs

guja-odoo:
bug:
Some icons in the Bistro preview were displayed as raw text. Its Material Symbols font URL still carried an ID.

fix:
Regenerate the preview so the font URLs are static.


bvr-odoo (approved):

bvr-odoo:
@robodoo r+

PR #1338, opened 2026-08-27
[FIX] themes: state illustration and device-mockup requirements

guja-odoo:
Descriptions didn't state that these themes use illustrations or device mockups instead of photos, so the AI recommender suggested them for shops, makers of physical goods, and businesses with no screen to show.

task-6456379


FrancoisGe:
@robodoo r+

PR #1345, opened 2026-09-11
[FIX] theme_eclipse: fix shaped icons in preview

bvr-odoo:
Steps to reproduce:
- Open the website configurator.
- Select a dark color palette.
- Preview the "Eclipse" theme. => The shaped icons in the Features section use a dark foreground.

Before this commit, the static preview applied its default dark icon color even when a background color class was present.

After this commit, the default icon color is limited to shaped icons without a background class, allowing `bg-o-color-X` to control their foreground.

task-6485048


sobo-odoo (approved):
@robodoo r+

